PR #14, opened 2020-08-06
Fix using a custom I18n scope

rbague:
Now, if you define a custom i18n_scope in your notification class it is not used, and noticed tries to find the translations under the default scope `notifications`. This PR fixes this, and adds a test for it so it does not break in the future.


excid3:
I was planning on adding this, thanks for beating me to it! 😄

PR #15, opened 2020-08-06
Allow to pass multiple params

rbague:
First of all, thanks for the awesome gem, something like this was really missing.

Right now, if you want to add multiple required params to a notification, you have to specify each of them in a separate line. With this PR, we can now specify multiple params in a single line, like this:
```ruby
class NewComment < Noticed::Base
  params :user_id, :comment_id
end
```

### DIFF
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -29,9 +29,10 @@ module Noticed
         new(params)
       end
 
-      def param(name)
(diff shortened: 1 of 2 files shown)


excid3:
This will break `param` if anyone was using that so far. Just want to make a note of that.

rbague:
That was fast, thanks!

PR #18, opened 2020-08-07
Fix: Make recipient available to 'user preference' check methods

patmisch:
Per the readme, `recipient` should be available in `if:` and `unless:` options. This doesn't seem to be the case in practice.

This should fix it. However, not sure if this is the best way to handle it or not.


excid3:
Thanks @patmisch! I believe that broke when I refactored to allow multiple recipients.

excid3:
We should get a test written for this too.

PR #23, opened 2020-08-10
Namespace the translation concern

rbague:
The translation concern should also be under the `Noticed` namespace to avoid conflicts


excid3:
Yes indeed, thanks @rbague 🙏

PR #29, opened 2020-08-14
Add scopes for 'read' and 'unread' notifications

kylekeesling:
Figured this might be a handy edition, and since it was so easy to add I figured I'd pass it along as a PR.


excid3:
Thank ya sir!

PR #32, opened 2020-08-23
Add generator for generating custom delivery methods

manojmj92:
Hello 👋 ,

This change adds a new generator, `noticed:delivery_method`, which helps to create a custom Delivery method class.

### Usage

To generate a new custom delivery method, simply run:

`rails generate noticed:delivery_method Name`

This generates a new delivery method class, under `app/notifications/delivery_methods`.

Eg: `rails generate noticed:delivery_method Discord`

### Other changes

I have also taken the liberty to enforce a clear distinction between naming `notifications` & `delivery_methods` in the docs.

For eg: The `Discord` delivery method class was previously named as `DiscordNotification`, while the `Notification` suffix was **also** being used to name `Notification` classes, like `CommentNotification`, which could have lead to some confusion for anyone reading the docs for the first time.

So, as a general rule the docs now enforces:

- Naming any notification class like `<Name of the subject>Notification`
- Naming any custom delivery class like `DeliveryMethods::<Name of the Service>`

Also, custom delivery method classes are now generated & namespaced under `DeliveryMethods`, similar to the naming convention used in the gem. 🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Unreleased
 
+* [NEW] Add `noticed:delivery_method` generator to create custom delivery methods.
+
 ### 1.2.11
 
 * [FIX] Use ActiveRecord configuration to detect adapter without establishing a database connection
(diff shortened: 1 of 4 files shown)


excid3:
Thanks @manojmj92! I don't know why I didn't think of making a delivery method generator yet! 😜

Consistency of naming is great and a very welcomed improvement as well. This looks fantastic.

PR #33, opened 2020-08-24
Validate that delivery by emails (`deliver_by :email`) always have a mailer specified.

manojmj92:
Hello 👋,

### Background

Our [code](https://github.com/excid3/noticed/blob/e115438b64ff0c181f1edefa11c7fc92e1f2f6a8/lib/noticed/delivery_methods/email.rb#L11) requires us to explicitly mention a `mailer` class when delivering notifications via `email`

ie, `deliver_by :email, mailer: 'UserMailer'`.

However, no validation currently exists to make sure that this has been specified correctly.

It is worth noting that the current code makes the email delivery fail (and just the email delivery, other channels mentioned in the notification class will work fine) anyway since the `mailer` attribute is not mentioned and errors out, but it fails with a `KeyError`, due to the usage of `fetch`, which is not a very friendly and informative error message to someone who is just starting out with this gem.

### Change

Without a `mailer` class explicitly specified, the `Notification` class does not start delivery thru any of the channels.

This is because it now fails with a `Noticed::ValidationError` before the processing of any channel starts, with a much more developer-friendly error message of `:email delivery method should specify a mailer`.

I have also added examples in the docs for `email` delivery methods so that `mailer` is explicitly mentioned everywhere. This is currently missing in the docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### Unreleased
 
+* [NEW] Validate that delivery by emails (`deliver_by :email`) always have a mailer specified.
 * [NEW] Allow validating options in custom delivery methods
 
 ### 1.2.12
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ To add delivery methods, simply `include` the module for the delivery methods yo
 class CommentNotification < Noticed::Base
   deliver_by :database
   deliver_by :action_cable
-  deliver_by :email, if: :email_notifications?
+  deliver_by :email, mailer: 'CommentMailer', if: :email_notifications?
 
   # I18n helpers
   def message
@@ -120,7 +120,7 @@ Like ActiveRecord, notifications have several different types of callbacks.
 ```ruby
 class CommentNotification < Noticed::Base
   deliver_by :database
-  deliver_by :email
+  deliver_by :email, mailer: 'CommentMailer'
 
   # Callbacks for the entire delivery
   before_deliver :whatever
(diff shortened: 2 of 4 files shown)


excid3:
This seems good! One of the things Rails 6.1 fixes is the ability to serialize a constant. 

The `mailer` option must be a string in Rails 6.0, but you'll be able to say `mailer: UserMailer` in 6.1 which is nice and kind of addresses this issue.

I like the idea of expanding the validations so we can validate other options. 👍

lib/noticed/base.rb
```diff
@@ -111,13 +111,28 @@ def get_class(name, options)
       end
     end
 
-    # Validates that all params are present
     def validate!
+      validate_params_present!
+      validate_email_delivery_methods_have_mailers_present!
```
excid3, line 116: Architecture question... should we loop through each delivery method and let them define validations?

I don't feel like the Notification base class is the right place to validate the mailer options, you know? Seems like it would make more sense to live in the `DeliveryMethods::Email` class. 

We could pass in the options to a class method on the delivery method for validation?

```
module DeliveryMethods
  class Base
    def self.validate!(options)
      # Override to raise ValidationError if passed options are invalid
    end
  end
end
```

class Noticed::Base
  def validate!
    delivery_methods.each do |method|
       # convert to class and call klass.validate!(method[:options])
    end
  end
end

excid3: Related https://github.com/excid3/noticed/pull/34

manojmj92: I ❤️  this idea, let me rework on this 🙂

excid3:
Just merged the other PR, so I think we're good to go once we get this updated to use the new `self.validate!(options)` method. 👍

manojmj92:
@excid3 thank you, this is ready for review now 🙂

excid3:
Thanks @manojmj92! This turned out nice and clean!

PR #34, opened 2020-08-24
Remove comment, tweak method

liroyleshed:


lib/noticed/base.rb
```diff
@@ -97,17 +97,16 @@ def run_delivery_method(delivery_method, recipient:, enqueue:)
...
       run_callbacks delivery_method[:name] do
-        klass = get_class(delivery_method[:name], delivery_method[:options])
+        klass = get_class_for(delivery_method[:name], delivery_method[:options])
         enqueue ? klass.perform_later(args) : klass.perform_now(args)
       end
     end
 
-    # Retrieves the correct class for a delivery method
-    def get_class(name, options)
+    def get_class_for(delivery_method, options)
```
excid3, line 105: Maybe we should call this `delivery_method_for(name, options)` ?

liroyleshed: I like it--made some changes.

PR #38, opened 2020-08-25
Allow validating options in custom delivery methods

manojmj92:
Based on the discussion that happened in https://github.com/excid3/noticed/pull/33, raising this MR to add the feature that allows to specify validations for options in custom delivery classes.

After this is merged (hopefully), the work in #33 can be completed by extending this ability to the `DeliveryMethods::Email` class.

The details of the usage has been provided in the README. 🙂

### DIFF
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -110,12 +110,24 @@ module Noticed
       end
     end
 
-    # Validates that all params are present
     def validate!
+      validate_params_present!
+      validate_options_of_delivery_methods!
+    end
+
+    # Validates that all params are present
+    def validate_params_present!
       self.class.param_names.each do |param_name|
         if params[param_name].nil?
           raise ValidationError, "#{param_name} is missing."
         end
+      end
+    end
+
+    def validate_options_of_delivery_methods!
+      delivery_methods.each do |delivery_method|
+        method = delivery_method_for(delivery_method[:name], delivery_method[:options])
+        method.validate!(delivery_method[:options])
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Unreleased
 
+* [NEW] Allow validating options in custom delivery methods
+
 ### 1.2.12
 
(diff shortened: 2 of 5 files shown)


lib/noticed/base.rb
```diff
@@ -110,13 +110,25 @@ def delivery_method_for(name, options)
...
         if params[param_name].nil?
           raise ValidationError, "#{param_name} is missing."
         end
       end
     end
+
+    def validate_options_of_delivery_methods!
+      delivery_methods.each do |delivery_method|
+        method = delivery_method_for(delivery_method[:name], delivery_method[:options])
+        method.validate!(delivery_method[:options])
```
excid3, line 130: I just noticed, `delivery_method_for` takes options as the second argument. We probably don't need that since it's just returning the class, not instantiating it. That would make this a bit more clear.

What if we did...

```
method = delivery_method_for(delivery_method[:name])
method.validate!(delivery_method[:options])
```

manojmj92: @excid3 I am sorry, but I am not sure if I understand.

`delivery_method_for ` currently requires the options incase we specify a custom `class` (in cases like custom delivery method)

```ruby
  deliver_by :discord, class: 'DeliveryMethods::Discord'
```

```ruby
    def delivery_method_for(name, options)
      if options[:class]
        options[:class].constantize
      else
        "Noticed::DeliveryMethods::#{name.to_s.classify}".constantize
      end
    end
```

So without passing in the `options`, we would not be able to figure out the class to be used?

I think the best we can do here is, instead of passing the whole options, we can just pass in `options[:class]`, so like,

```ruby
  method = delivery_method_for(delivery_method[:name], delivery_method[:options][:class])
  method.validate!(delivery_method[:options])
```

Anything I am missing? 🙂

excid3: Ah yeah, you're right. Ignore me. We may want to revisit that in the future and refactor.

manojmj92: @excid3 no worries, thanks for clarifying 🙂

PR #47, opened 2020-09-18
Delivery method option validations

rbague:
Adds an api to validate deliver method options presence. It uses the same api as the notification params, to keep everything consistent.

It uses the `validate!` method from the `Noticed::DeliveryMethods::Base` class, so it does not break any backwards compatibility.

```ruby
class DeliveryMethods::Discord < Noticed::DeliveryMethods::Base
  option :username # Requires the username option to be set

  def deliver
    # Discord delivery logic
  end
end
```

### DIFF
--- a/lib/noticed/delivery_methods/base.rb
+++ b/lib/noticed/delivery_methods/base.rb
@@ -22,7 +22,7 @@ module Noticed
 
         def validate!(delivery_method_options)
           option_names.each do |option_name|
-            if delivery_method_options[option_name].nil?
+            unless delivery_method_options.key? option_name
               raise ValidationError, "option `#{option_name}` must be set for #{name}"
             end
           end
--- a/lib/noticed/delivery_methods/base.rb
+++ b/lib/noticed/delivery_methods/base.rb
@@ -4,7 +4,30 @@ module Noticed
       extend ActiveModel::Callbacks
       define_model_callbacks :deliver
 
+      class_attribute :option_names, instance_writer: false, default: []
+
       attr_reader :notification, :options, :params, :recipient, :record
+
+      class << self
+        # Copy option names from parent
+        def inherited(base) #:nodoc:
+          base.option_names = option_names.dup
+          super
+        end
+
+        def options(*names)
+          option_names.concat Array.wrap(names)
+        end
+        alias option options
+
+        def validate!(delivery_method_options)
+          option_names.each do |option_name|
+            if delivery_method_options[option_name].nil?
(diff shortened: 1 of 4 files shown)


lib/noticed/delivery_methods/base.rb
```diff
@@ -4,8 +4,31 @@ class Base < Noticed.parent_class.constantize
...
+        end
+
+        def options(*names)
+          option_names.concat Array.wrap(names)
+        end
+        alias option options
+
+        def validate!(delivery_method_options)
+          option_names.each do |option_name|
+            if delivery_method_options[option_name].nil?
```
manojmj92, line 25: Question: I am wondering if we should avoid going to the length of making this akin to `validates_presence_of` validator like in Rails. In the previous implementation, we only check if the `key` is present with this specific name, and not the value of the item itself.

I wonder if checking if the value itself is `nil` could be troublesome at least to some users of the gem, in case there is a legitimate requirement to pass `nil` as the value of an option or if the value is set by an ActiveRecord query which may return `nil` in some cases, say like `also_send_to: User.find_by(id: admin_id)`.

Or is it deliberate that you wanted to make this a presence validator? Would like to know your thoughts 🙂

manojmj92 (commented):
Thanks, @rbague, love this change and it is super neat ❤️. I have one question for your consideration 🙂

lib/noticed/delivery_methods/base.rb
rbague: You are totally right, this would break backwards compatibility. The reason of checking for the options presence, is that it is the same validation used for notifications [params](https://github.com/excid3/noticed/blob/ed0527763c3bf1a326054ac98080716136ca5396/lib/noticed/base.rb#L121).
I agree with you that it will make more sense to only check if the key is present in the options hash. But before making the change, I'd like to know @excid3 opinion on it, because we may also want to apply this change to the params validation, so everything is consistent.

excid3: I think it should only check for the presence of the option, not the value. We don't really care what the value is because that's up to the delivery method to decide, so long as something was passed in.

excid3:
I like the idea of applying this to delivery methods. 👍

I think we should just make sure it only validates that an option was passed in (rather than the value) and that's it.

excid3 (changes requested):
I like the idea of applying this to delivery methods. 👍

I think we should just make sure it only validates that an option was passed in (rather than the value) and then it's good to go.

excid3:
Thanks @rbague 🙏

PR #51, opened 2020-10-15
Add index to read_at column in generator

mikelkew:
The `read_at` attribute is used in a number of scopes, so it is preferable to have an index on it to improve query performance.

This is a quick PR to add an index on `read_at` by default when running the generator.


excid3:
Definitely forgot to add that. Thanks @mikelkew 🙏

PR #52, opened 2020-10-16
Use correct adapter name for detecting MySQL

mikelkew:
Quick PR to ensure that `json` is used by MySQL for the `params` column in the generator.

Previously it would use `text` because the case statement condition was `mysql`, but the correct MySQL adapter name is `mysql2`.


excid3:
Thanks a bunch @mikelkew 🙏

Issue #54, opened 2020-10-19
Mark individual notifications as unread


pdunleav:
Hello!

Love the gem! 🙌 🙂.   Just had one instance method I found myself adding to the `Notification` model that may be a candidate for inclusion - to mark an individual notification as unread again if it can be saved for later. 

```ruby
def mark_as_unread!
  update(read_at: nil)
end
```

Is this something that would suit being added to `model.rb`, or perhaps this was something that was not done on purpose for a reason I am unaware of 😅. If it is something that could be added I am happy to submit the PR, just wanted to check for suitability/interest first! Thanks again for the awesome gem 🙌

excid3:
Great point @pdunleav! I hadn't had any need for marking notifications unread yet so it didn't cross my mind. Makes perfect sense to add to the gem I think. Send over a PR! 💪

excid3:
Closed by #56

PR #56, opened 2020-10-19
add instance method to mark notification unread

pdunleav:
this adds enhancement from issue https://github.com/excid3/noticed/issues/54


excid3:
Thanks so much @pdunleav 🙏

PR #57, opened 2020-10-26
Added support for Microsoft Teams

jordanbrock:
Added a MS Teams delivery method, and associated test.

I've basically just copied the Slack delivery method, and added some documentation about how to structure the payload for an MS Teams card. 

Please let me know what you need me to do to make this ready to be merged.

### DIFF
--- a/lib/noticed.rb
+++ b/lib/noticed.rb
@@ -16,6 +16,7 @@ module Noticed
     autoload :Database, "noticed/delivery_methods/database"
     autoload :Email, "noticed/delivery_methods/email"
     autoload :Slack, "noticed/delivery_methods/slack"
+    autoload :Team, "noticed/delivery_methods/team"
     autoload :Test, "noticed/delivery_methods/test"
     autoload :Twilio, "noticed/delivery_methods/twilio"
     autoload :Vonage, "noticed/delivery_methods/vonage"
--- a/lib/noticed.rb
+++ b/lib/noticed.rb
@@ -16,7 +16,7 @@ module Noticed
     autoload :Database, "noticed/delivery_methods/database"
     autoload :Email, "noticed/delivery_methods/email"
     autoload :Slack, "noticed/delivery_methods/slack"
-    autoload :Team, "noticed/delivery_methods/team"
+    autoload :MicrosoftTeam, "noticed/delivery_methods/microsoft_team"
     autoload :Test, "noticed/delivery_methods/test"
     autoload :Twilio, "noticed/delivery_methods/twilio"
     autoload :Vonage, "noticed/delivery_methods/vonage"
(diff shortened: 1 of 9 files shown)


excid3 (changes requested):
This looks great @jordanbrock! The only change I have for you is can we renaming the delivery method to `:microsoft_teams` and change the class and filename accordingly?

lib/noticed/delivery_methods/team.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      def format
+        if (method = options[:format])
+          notification.send(method)
+        else
+          {
+            title: notification.params[:title],
+            text: notification.params[:text],
+            sections: notification.params[:sections],
+            potentialAction: notification.params[:notificationAction]
```
rbague, line 18: Shouldn't the `notificationAction` key be underscored? That way it would be consistent with the rest of the keys in the project
```suggestion
            potentialAction: notification.params[:notification_action]
```

rbague (commented):
Hi @jordanbrock, thanks for the addition. I just have one question regarding key naming

lib/noticed/delivery_methods/team.rb
rbague: Yes, I totally agree. It was just to keep the public API as consistent as possible

excid3: As long as the other keys don't need changing too much, it should be a good improvement. 👍

jordanbrock:
I've changed the delivery method name to :microsoft_teams and incorporated @rbague's fix for the key name.

Let me know if there's anything else?

jordanbrock (commented):
Changes made as requested

lib/noticed.rb
```diff
@@ -16,6 +16,7 @@ module DeliveryMethods
     autoload :Database, "noticed/delivery_methods/database"
     autoload :Email, "noticed/delivery_methods/email"
     autoload :Slack, "noticed/delivery_methods/slack"
+    autoload :MicrosoftTeam, "noticed/delivery_methods/microsoft_team"
```
excid3, line 19: Sorry to be nitpicky. Can we pluralize this since Microsoft calls it "Teams" plural?

excid3, line 19: This will need to be plural to match the delivery method key. Plus, Microsoft calls it "Teams", so we should make sure it matches. Sorry to be nitpicky! 😅

jordanbrock: I have tried doing that initially, but I ended up with an error with the class not being found. I'll have another crack in case it was something else causing it

excid3: Looks like the `classify` in the run_delivery_method singularizes it. Might be able to just camel case it in some other way that doesn't lose the pluralization.

`"Noticed::DeliveryMethods::#{name.to_s.classify}".constantize`

excid3: Looks like `camelize` instead of `classify` will do what we want. 👍 

Want to change that too?

jordanbrock: Shall do.

excid3: Appreciate it!

excid3 (approved):

PR #59, opened 2020-10-30
Ensure Vonage delivery returns an HTTP::Response

### LINKED ISSUE #53: Get response from delivery method in callback
I wonder if it's currently possible, or whether it would be a useful feature to be able to access the delivery response in an after callback.

```rb
class InvitationNotification < ApplicationNotification
  deliver_by :vonage

  after_vonage :record_vonage

  # ...

  private 

  def record_vonage
    recipient.deliveries.create(status: vonage_response)
  end
end
```

This is a contrived example (using `vonage_response`, which would be the status code from [this line](https://github.com/excid3/noticed/blob/master/lib/noticed/delivery_methods/vonage.rb#L6). This would provide the flexibility to actually record whether a delivery method was successful or not.

My specific use-case is being able to determine whether text messages where delivered or not as they're displayed back to a user, though I'm sure it would be equally beneficial for most delivery methods if you needed logging/tracking of the response.

rbague:
Closes #53 by ensuring the Vonage delivery method returns the HTTP response object in its delivery method, so it can be used in an `around` callback (as explained [here](https://github.com/excid3/noticed/issues/53#issuecomment-717886090)).

I've also added tests for all the delivery methods so that any new changes don't break this "feature". In the end, we'll probably want to standardize the response of the delivery action of all delivery methods (as @excid3 suggested [here](https://github.com/excid3/noticed/issues/53#issuecomment-717958629)).

### DIFF
--- a/lib/noticed/delivery_methods/email.rb
+++ b/lib/noticed/delivery_methods/email.rb
@@ -4,7 +4,7 @@ module Noticed
       option :mailer
 
       def deliver
-        mailer.with(format).send(method.to_sym).deliver_later
+        mailer.with(format).send(method.to_sym).deliver_now
       end
 
       private
--- a/lib/noticed/delivery_methods/vonage.rb
+++ b/lib/noticed/delivery_methods/vonage.rb
@@ -7,6 +7,8 @@ module Noticed
         if !options[:ignore_failure] && status != "0"
           raise ResponseUnsuccessful.new(response)
         end
+
+        response
       end
 
       private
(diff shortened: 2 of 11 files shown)


rbague:
Just for mentioning it, whilst working on this, I realized that the `ActionCable` and `Email` delivery methods return "strange" values.
The ActionCable delivery method returns the response of broadcasting though action_cable, which is `nil` (I don't know if it would make more sense to explicitly return nil, for clarity, and in case the return value of broadcast changes in the future).
And the Email delivery returns the scheduled active_job that will send the email later on (which I don't know if it is useful at all to the user, and it may be better to return nil)

excid3:
Looking good @rbague 👍 

ActionCable - I don't think the return value matters much on that one.
Email - I actually don't know that we need to `deliver_later` here since we're already in a background job. It's redundant to queue up another job. What if we change it to `deliver_now` and use that return value?

rbague:
It does make more sense to deliver the email the same moment the job is executed, instead of triggering a new one. Which also solves the "issue" of returning the mailer job, and now returns a `Mail::Message` object, which can be useful for debugging

excid3:
This looks great. Thanks for the help on this @rbague!

PR #61, opened 2020-11-05
Add option to delay the delivery

### LINKED ISSUE #50: Checking if notification is read before sending email/text
Hi, I'm not sure if there is a better way to do this, but I wanted to see if there is a standard way to delay the deliver_by email/twilio methods and check if the notification is read off the database before sending. Right now, I have it built into my conditional check in the delivery method.
```
deliver_by :email, mailer: "UserMailer", method: :message_notification, if: :email_notifications?

def email_notifications?
  sleep(60)
  record.reload
  if record.read?
    return false
  else
...
```

rbague:
Add an option to delay the delivery of a delivery method. To use it, just add the `delay` option with a valid duration
```ruby
class CommentNotification < Noticed::Base
  deliver_by :database
  deliver_by :email, mailer: 'CommentMailer', delay: 5.minutes, unless: :read?

  def read?
    record.read?
  end
end
```

I've had to create a new job instead of using the delivery method one so the `if` and `unless` options run after the given delay, otherwise they would be triggered the moment the notification is delivered.

Closes #50


excid3:
Ah shoot, the `if` and `unless` do make that more complex don't they... hmm. I was hoping we wouldn't have to do that, but I guess it makes sense.

excid3:
Do you think there are any other alternatives we could do to keep things simpler? 

I could see putting the `if` and `unless` as the first steps of the delivery method jobs? That way we'd always queue up jobs, they just may not do anything. It would keep things simpler in theory?

rbague:
Yes, if we move the `if` and `unless` options to the delivery_method the implementation would be much simpler, and easier to read/understand. I don't see any other simple alternative.

The only downside I see is the one you pointed, we'd be always queuing up jobs even if they do nothing. I guess it is acceptable as it shouldn't affect performance much.

If you agree on this I can have the change made in no time

excid3:
Let's do that. It simplifies a lot and will only queue up a small handful of extra jobs that will run almost instantly.

It's normal for jobs to be designed like that where the job determines if it should run or not. Perham recommends it, especially if you're queuing up jobs you might need to cancel. 

For anyone who comes across this thread: this adds a feature where you can delay the email by 5 minutes and use `if` and `unless` in the job to see if the notification was already marked as read and cancel the email. 👍

rbague:
Awesome! Just pushed the changes.

excid3:
Slick, that turned out great. 💪

gathuku:
Is it possible to pass the delay time from the params? something like

```
CommentNotification.with(comment: "some comment", delay: 3.minutes).deliver_later(User.first)
```

```
class CommentNotification < Noticed::Base
  deliver_by :database
  deliver_by :email, mailer: 'CommentMailer', delay: params[:delay]

end

```

excid3:
No way to do that currently. I think we could add support for that by allowing you to pass a proc / lambda for the option which we could evaluate later.

Do you have an example use case?

gathuku:
In my use case am using a custom delivery to send SMS, I wanted a way where you can schedule an SMS to be sent later(at a specific dateTime). This means you will pick dateTime when you want the SMS to be delivered. What I tried is to get the chosen DateTime difference with `Time.now` in minutes pass it to my notification instance as params and finally set it as `delay` in the delivery method.

rbague:
If I understand correctly, what you are trying to achieve is not to delay the notification, but schedule it to a given time. Am I right?
If that's the case, I think it could be a great addition, and something pretty easy to implement without adding a lot of complexity to the library. What do you think @excid3 ?

Something like the following could solve your issue @gathuku ?
```ruby
CommentNotification.with(comment: "some comment").at(Date.tomorrow.noon).deliver(User.first)
```

gathuku:
Yes @rbague you are right!

excid3:
Moved this discussion to #63

PR #62, opened 2020-11-09
Use WebMock to stub http requests in testing

### LINKED ISSUE #60: Record HTTP requests for testing
Whilst working on #59 I noticed that the `Vonage`, `Slack`, `Twilio` and `MicrosoftTeams` tests trigger a real HTTP request to different services, which always fail and raise a `Noticed::ResponseUnsuccessful` exception, making all of those tests a little bit more complicated to read and write.

It may be great if we could locally record successful requests to each service, and use them in the tests, so we could replicate real case scenarios. I've used [WebMock](https://github.com/bblimke/webmock) in the past for this specific case and it works great.

rbague:
Install and configure [WebMock](https://github.com/bblimke/webmock) to prevent sending real HTTP requests, and test real-case scenarios.

With those changes, the dependency on [mocha](https://github.com/freerange/mocha) can be removed as it is no longer used in the test files

Closes #60

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -69,6 +69,7 @@ GEM
     builder (3.2.4)
     byebug (11.1.3)
     concurrent-ruby (1.1.7)
+    crack (0.4.4)
     crass (1.0.6)
     domain_name (0.5.20190701)
       unf (>= 0.0.5, < 1.0.0)
@@ -79,6 +80,7 @@ GEM
       rake
     globalid (0.4.2)
       activesupport (>= 4.2.0)
+    hashdiff (1.0.1)
     http (4.4.1)
       addressable (~> 2.3)
       http-cookie (~> 1.0)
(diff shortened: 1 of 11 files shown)


rbague:
The only things I don't like are the tests that still trigger an HTTP request ([Microsoft Teams](https://github.com/rbague/noticed/blob/7b41a1ba850ce0990b33d3921bc9b23d94108a12/test/delivery_methods/microsoft_teams_test.rb#L41), [Slack](https://github.com/rbague/noticed/blob/7b41a1ba850ce0990b33d3921bc9b23d94108a12/test/delivery_methods/slack_test.rb#L20), [Twilio](https://github.com/rbague/noticed/blob/7b41a1ba850ce0990b33d3921bc9b23d94108a12/test/delivery_methods/twilio_test.rb#L24) and [Vonage](https://github.com/rbague/noticed/blob/7b41a1ba850ce0990b33d3921bc9b23d94108a12/test/delivery_methods/vonage_test.rb#L27)) so that we can test that a `Noticed::ResponseUnsuccessful` is raised if an error is found.

excid3:
Thanks for taking the stab at this @rbague!

I went ahead and removed mocha as well as added failure responses (just 403s) to simulate errors that way no tests are making HTTP requests.

PR #64, opened 2020-11-11
Database delivery cannot be delayed

rbague:
This PR ensures that the database delivery can't be delayed and is executed right away, so that the rest of the deliveries can access the database record.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Unreleased
 
+* [FIX] Database delivery can't be delayed, otherwise the database record won't be available for the other deliveries - @rbague
+
 ### 1.2.18
 
 * [NEW] Add `delay` option to delay the delivery of a specific delivery method - @rbague
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ Writes notification to the database.
 
 `deliver_by :database`
 
-**Note:** Database notifications are special in that they will run before the other delivery methods. We do this so you can reference the database record ID in other delivery methods. For that same reason, this delivery can't be delayed, so the `delay` option has no effect in here.
+**Note:** Database notifications are special in that they will run before the other delivery methods. We do this so you can reference the database record ID in other delivery methods. For that same reason, the delivery can't be delayed (via the `delay` option) or an error will be raised.
 
 ##### Options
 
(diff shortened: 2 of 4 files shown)


rbague:
I've also added a couple of comments to describe why database delivery is a special case. I can remove them if you think the code is self-explanatory.

lib/noticed/base.rb
```diff
@@ -74,7 +74,13 @@ def run_delivery(recipient, enqueue: true)
 
       # Run database delivery inline first if it exists so other methods have access to the record
       if (index = delivery_methods.find_index { |m| m[:name] == :database })
+        # Remove it from the array of deliveries so it is only executed once
         delivery_method = delivery_methods.delete_at(index)
+
+        # If delayed, run_delivery_method will return the scheduled job instead of the record
+        delivery_method[:options].delete(:delay) if delivery_method[:options].key?(:delay)
```
excid3, line 81: What if we do this as a validation on the DeliveryMethods::Database class instead? That way we can raise an `ArgumentError` with a message about it right there.

test/noticed_test.rb
```diff
@@ -167,4 +172,12 @@ class Noticed::Test < ActiveSupport::TestCase
       With5MinutesDelay.new.deliver(user)
     end
   end
+
+  test "database delivery is executed without a delay" do
```
excid3, line 176: If we do the validation, we can have a test to catch the `ArgumentError` raised instead.

excid3 (changes requested):
Thanks for this! Definitely makes sense. I left a couple comments because I think we can maybe organize this slightly better as a validation on the database delivery method instead. Let me know what you think!

test/noticed_test.rb
rbague: I'll also move it to the `test/delivery_methods/database_test.rb` file, as it makes more sense to be there

excid3: 👍

rbague:
I've also added a test to ensure that the database delivery is not enqueued and executed right away, just to ensure this keeps working and doesn't break as I did in #61

excid3 (approved):
Looks great! Only small tweak I did was call `super` in the validate method to make sure any default validations still run for it.

rbague:
Thank you, I totally forgot it! 🙏

excid3:
Released 1.2.19 👍

PR #67, opened 2020-11-26
Delay Entire notifications

### LINKED ISSUE #63: Add delay to entire notification
We just added delays for individual delivery methods which is great (thanks @rbague!). 

@gathuku mentioned it would be great to delay the entire notification: https://github.com/excid3/noticed/pull/61#issuecomment-724941484

```
In my use case am using a custom delivery to send SMS, I wanted a way where you can schedule an SMS to be sent later(at a specific dateTime). This means you will pick dateTime when you want the SMS to be delivered. What I tried is to get the chosen DateTime difference with Time.now in minutes pass it to my notification instance as params and finally set it as delay in the delivery method.
```

I think this makes sense to do and think we could probably allow similar syntax as ActiveJob:

```
Job.set(wait: 1.week).perform_later
Job.set(wait_until: Date.tomorrow.noon).perform_later
```

We should support `.set(**options)` syntax and simply pass those along on a `deliver_later` call. This would keep the interface nicely similar to ActiveJob and flexible.

gathuku:
Delay entire notification with an active job like syntax

```
NewComment.with(comment: @comment).set(wait: 20.minutes).deliver_later(user)
```
Pass other rails supported options `wait_until`, `queue`, `priority` 

Closes #63

### DIFF
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -8,9 +8,10 @@ module Noticed
 
     class_attribute :delivery_methods, instance_writer: false, default: []
     class_attribute :param_names, instance_writer: false, default: []
+    class_attribute :set_options, default: {}
 
     # Gives notifications access to the record and recipient when formatting for delivery
-    attr_accessor :record, :recipient, :set_options
+    attr_accessor :record, :recipient
 
     class << self
       def deliver_by(name, options = {})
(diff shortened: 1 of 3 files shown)


gathuku:
@excid3 what could be the best approach to this feature? please help we make it better.

stayce:
Any updates? This would be a great feature addition

excid3:
This is now supported in v2.

PR #71, opened 2020-12-05
Add docs for fallback / escalating notifications

### LINKED ISSUE #70: Strategy for fallback notifications?
Would be great to support (or document) how to handle this common use-case:

* Notify a user via `database` / in-app notification
* If user has not marked notification as read / seen after a certain time, notify via `email` or other channel.

swanson:
Resolves #70


excid3:
This is wonderful. Thanks @swanson

PR #73, opened 2020-12-07
Merge recipient and record in both default and custom formatting method

### LINKED ISSUE #72: Documentation is not clear
https://github.com/excid3/noticed/blob/d147d49591b38a141b24634fdcbafa4b5e7641f9/lib/noticed/delivery_methods/email.rb#L22

This line is not merging the recipient

https://github.com/excid3/noticed/blob/master/README.md#L220

> Use a custom method to define the params sent to the mailer. recipient will be merged into the params.

I had to merge it myself.

silva96:
In order to be consistent with the documentation, merge recipient and record in both default and custom formatting branches inside the format method.

closes #72

### DIFF
--- a/lib/noticed/delivery_methods/email.rb
+++ b/lib/noticed/delivery_methods/email.rb
@@ -18,14 +18,12 @@ module Noticed
       end
 
       def format
-        if (method = options[:format])
+        params = if (method = options[:format])
           notification.send(method)
         else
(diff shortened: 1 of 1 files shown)


excid3:
Thanks @silva96 👍 I have no idea why I didn't do that in the first place. Makes much more sense this way.

PR #75, opened 2020-12-10
Run bundle exec appraisal rake test

MatthewKennedy:
I might be incorrect, but I don't think the spec tests were being run, because ```bundle exec appraisal install``` was being run in place of ```bundle exec appraisal rake test```.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,7 @@ jobs:
           DATABASE_URL: mysql2://root:password@127.0.0.1:3306/test
           RAILS_ENV: test
         run: |
+          bundle exec rails db:create
           bundle exec appraisal rake
 
 
@@ -145,4 +146,5 @@ jobs:
         env:
           RAILS_ENV: test
         run: |
+          bundle exec rails db:create
           bundle exec appraisal rake
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,8 @@ on:
     branches:
       - master
 jobs:
+
+  # MySQL ###
   mysql:
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


excid3:
There's an open issue with Appraisal right now that I've tried fixing and I think I had just left this half finished (and forgot about it). 

I'm not entirely sure why this failed saying it couldn't find the appraisal gem.

excid3:
Not using my branch in ActsAsTenant and it seems to be working fine: https://github.com/ErwinM/acts_as_tenant/blob/master/.github/workflows/ci.yml

Maybe we just need to move back to the appraisal gem from Rubygems.

MatthewKennedy:
@excid3 I'm looking into this on a separate rails gem of mine, so far I've got Postgres & Sqlite working, just testing MySQL now and then test them all in the same file.

If it all works I'll move the yml code into this branch.

MatthewKennedy:
@excid3 Postgress tests pass, but fail for others:

```rails aborted!
NoMethodError: undefined method `jsonb' for #<ActiveRecord::ConnectionAdapters::MySQL::TableDefinition:0x00005608a0076d28>```

```
* [NEW] Update generator to make params column json for MySQL, jsonb for Postgres, and text for everything else
--
57 | * [FIX] Keyword args warning for delivery methods is now fixed
… |  
80 | * [FIX] Improve serializer to handle text, json, and jsonb columns```

excid3:
@MatthewKennedy mysql only has a `json` column type, `jsonb` is for Postgres. Sounds like it's running the wrong column type in the migration.

We check the database adapter in the generator and use the type according to that: https://github.com/excid3/noticed/blob/master/lib/generators/noticed/model_generator.rb#L44

Could be a bug in there potentially, although it is passing on master?

PR #76, opened 2020-12-14
Set Up Ci for Multiple DB's

MatthewKennedy:
- Setup `ci.yml` ready for testing against multiple database types.
- Move Standard RB into its own workflow.
- Use cache@v2.
- Add StandardRB badge to Readme.


excid3:
@MatthewKennedy One thing I learned yesterday was that `ruby/setup-ruby` has built-in bundle install and cache support which is super interesting: https://github.com/ruby/setup-ruby#caching-bundle-install-automatically

And the libsqlite-dev package is already installed, so that can be skipped too. Got the test suite for acts_as_tenant to run in under 15 seconds for some things which is crazy fast.

https://github.com/ErwinM/acts_as_tenant/blob/master/.github/workflows/ci.yml

MatthewKennedy:
Hi @excid3 I was watching the PR's on acts_as_tenant, Artem Kozaev did some sterling work, I'll look into both the link you mentioned and the work on acts_as_tenant workflow folder.

### False Passing of Tests Run In Travis-CI
I have been working on a gem that I forked , it has separate ways of working for Postgres, MySQL and SQLite and the tests were setup to expect different results depending on the database in use, stuff like jsonb, and case insensitive results...

Anyhow, the tests were all passing in Travis-ci, so naturally when some of the Posgress tests failed on GitHub Actions my instinct was that something was wrong with the GitHub Actions setup, especially as these tests had been used long before I came along.

I decided to switch the local database.yml config (it was setup for SQLite), and replace the contents with those for a Postgres connection and re-run the tests locally, sure enough the tests failed, just like in GitHub Actions.

All along the tests that were passing in Travis-CI were giving a false pass, because the database.yml file in the dummy/configs had the SQLite connection details, so the spec tests specific to Postgress and MySQL were never being run.

### Ensure Correct Database Connection
- git ignore the dummy app database schema.
- Create a rake task that creates a new test database and migrates.
- Have the GitHub Actions swap the content of the database.yml for those specific to the database used in the current test, then run the new rake task to make sure the datable schema is correct for the direct encoding types.

See the link below, the ruby was looking for all types of database connection info and basing tests off the, all it was getting back in travis-ci was sqlite.

[link(https://github.com/MatthewKennedy/make_taggable/blob/master/spec/support/helpers.rb)

excid3:
We're using DATABASE_URL to force the database (skipping the database.yml) in GitHub Actions, so I don't think we've got the same issue as you're describing. That's the same way that Heroku, Hatchbox, etc work in production to set the DATABASE_URL for your production db.

MatthewKennedy:
That's explains why the tests were running properly in GitHub actions for Postgres, and not Travis-CI.

Strangely though the MySQL tests just wouldn't run, and was giving an error of incorrect encoding "unicode".

excid3:
Docker may not create the db with unicode by default? We're just using this: https://hub.docker.com/_/mysql

noticed.gemspec
```diff
@@ -19,8 +19,12 @@ Gem::Specification.new do |spec|
   spec.add_dependency "rails", ">= 6.0.0"
   spec.add_dependency "http", ">= 4.0.0"
 
+  spec.add_development_dependency "appraisal"
   spec.add_development_dependency "pg"
   spec.add_development_dependency "standard"
   spec.add_development_dependency "webmock"
   spec.add_development_dependency "mysql2"
+
+  spec.add_development_dependency "rspec", "~> 3.10.0"
```
excid3, line 28: Just a heads up, I definitely want to stick with just using the Rails test suite, not Rspec.

MatthewKennedy:
Hi @excid3

To get this gem to run the tests on other databases than just Postgres, I think I would have to create some migration files, and a rake task to create db, and run the migrations,  this would create a schema specific to the database type in use on each particular test.

Is this a good idea or a bad idea?

My thinking is the hard coded schema that seems specific to Postgres won't play nice with MySQL or SQLite.

PR #99, opened 2021-03-12
Change the semantics to have the classes that inherit from Noticed::Base called Notifiers

### LINKED ISSUE #98: Terminology Question/Suggestion
First and foremost, I want to say a big thank you for this library. It is awesome and rock-solid and has been seriously useful to my team. When I introduced this to the team, though, there was one point in particular that tripped them up and took some time to clarify: two things are called "Notification".

One of the one hand, you have the `SomeNotification` class that inherits from `Noticed::Base`; on the other hand, you have the `Notification` model which stores the individual database notification records. This terminology confusion is made completely explicit whenever you do something like this when trying to render database notifications in-app: `@notification.to_notification`.

In our team, we have started calling the Noticed classes "Notifiers", so our class names are now `SomeNotifier` and we aliased the `to_notification` method on the `Notification` model to `to_notifier`. This has helped our communication, so I wanted to open the floor for a discussion around the terminology. If you have some alternative terminology you think is better, I'm willing to do the PR either way.

fractaledmind:
@excid3: I manually went thru every single file to try my best to not change any reference to the database `Notification` class, but of course this should have a thorough set of second eyes. As you said in the issue, this is a breaking change so I bumped the version to 2.0.0. This also had the follow on effect of having the Rails version in the `Gemfile.lock` bumped as well to 6.1.3

resolves #98


fractaledmind:
I am confused by all of the GitHub Action failures. I don't see how anything I did would break things in the way they are breaking.

excid3:
This will be a big one! I definitely think it will help with the wording. 

I think if you drop the Gemfile.lock and version changes, that will help doing a merge. 👍

fractaledmind:
@excid3: Yep, got ahead of myself. Reverted those changes, merged in the most recent pushes to `master` and now everything is green

shamangeorge:
Hi! Any idea when this will be merged?

fractaledmind:
@excid3: Resolved conflicts with `master`.

nitsujri:
Oh man, I like this change a lot.

It semantically/mentally decouples the AR model from the "handlers".

I just started using this gem and based on the current semantics I thought the `CommentNotification` should have been a single table inheritance from `Notifications` esp since it uses the `type` column (clearly [it's not STI](https://github.com/excid3/noticed/blob/master/lib/noticed/model.rb#L14), but initial reading...).

Felt like I should be able to do `CommentNotification.where(recipient: user)` to easily get back a specific type of notification and was surprised I couldn't.

fractaledmind:
All good. I figured. Let me know if/when you want me to update the PR with head of main

fractaledmind:
@excid3: Updated against head of `master`

excid3:
I rolled this into v2. 🎉

PR #103, opened 2021-03-26
Add option for customizing a delivery method's queue

iheanyi:
By default, `noticed` uses whatever the default queue name is for `ActiveJob` for enqueuing things. At times, the user wants to use different queues for things, so we should be able to customize that queue on a per-delivery method basis when necessary.


excid3:
Love this, thanks @iheanyi 💪 

Looks like it's just failing on the mimemagic dependencie (what isn't right now? 😅). I think you can just update that with `bundle update && appraisal update` and then it should pass.

test/noticed_test.rb
```diff
@@ -167,4 +171,10 @@ class Noticed::Test < ActiveSupport::TestCase
       With5MinutesDelay.new.deliver(user)
     end
   end
+
+  test 'asserts delivery is queued with different queue' do
+    assert_enqueued_with(queue: 'custom') do
+      WithCustomQueue.new.deliver(user)
```
excid3, line 177: I think you need this to be `deliver_later` to fix your test. `deliver` won't enqueue anything cuz it runs immediately.

iheanyi:
@excid3 Welp, tests are failing and I don't know why, haha. So let me go back to the drawing board on this one and see if I can fix it.

test/noticed_test.rb
iheanyi: Oh, derp, good catch. Was copying off `delay`, but delay will always be done async.

iheanyi: Thanks!

iheanyi:
Okay, we're all green now, it looks like. Thanks for your help with this!

excid3:
No problem and thanks for the PR! 💪 

I'll probably cut a new release for it in just a minute.

excid3:
1.3.2 released 🚀

PR #105, opened 2021-03-27
Compatibility for Rails 5.2

### LINKED ISSUE #104: Add support for Rails 5.2
Many projects are on Rails 5.x, and because some level of support for ActiveJob and ActionCable is found in version 5.2, there shouldn't be too much fiddling around necessary to offer support for this version.  (And probably not so easy to go any further back to 5.1 or 5.0, but not as much need either since the folks on earlier 5.x versions can likely very easily upgrade their project to 5.2.)

lorint:
Resolves #104

Motivation:
  Many projects are on older versions of Rails, and because some level of support
  for ActiveJob and ActionCable are found in Rails 5.2, there isn't too much
  fiddling around that's necessary in order for Noticed to support this older environment.

Implementation:
  From Rails 6, port over the ActionCable test adapter and test helpers, and also the
  Serializer support in ActiveJob, along with the SymbolSerializer and DurationSerializer.
  And use a freeze_time block for the one time-sensitive test that expects an exact five
  minute interval to elapse before sending notification.

Sample test run of Noticed 1.3.2 running under Ruby 2.6.5 and Rails 5.2.5:

image


lorint:
Strangely all Github Actions test runs are currently failing during a very early step -- "Set up Ruby" -- with the message, "The process '/opt/hostedtoolcache/Ruby/2.7.2/x64/bin/gem' failed with exit code 2".
Screen Shot 2021-03-28 at 7 23 18 AM
All ideas are welcome!

Gemfile
```diff
@@ -14,3 +14,5 @@ gemspec
 # To use a debugger
 gem "byebug", group: [:development, :test]
 gem "appraisal", github: "excid3/appraisal", branch: "fix-bundle-env"
+# Temporary patch for the big mimemagic SNAFU
+gem "mimemagic", github: "mimemagicrb/mimemagic", ref: "01f92d86d15d85cfd0f20dabd025dcbd36a8a60f"
```
excid3, line 18: Now that the new Rails releases are out, I think you can drop this and update the Rails versions: https://weblog.rubyonrails.org/2021/3/26/marcel-upgrade-releases/

excid3:
I'm fine with adding 5.2 support even though Rails 5.2 probably will not be maintained very much longer though with Rails 7 slated for release around Railsconf 2021 and they only support the 3 latest versions (iirc).

lorint:
> (fine with adding 5.2 support)

We do appreciate this as it's a key deciding factor in being able to utilise this gem.  Thank you kindly!

Gemfile
lorint: Ah yes indeed -- we are green again now.
Important to have the rails_master lock to be happy was to remove mini_portile2 :)

test/delivery_methods/database_test.rb
```diff
@@ -17,7 +17,7 @@ class WithDelayedDatabaseDelivery < Noticed::Base
     end
 
     assert_equal 1, user.notifications.count
-    assert_equal :bar, user.notifications.last.params[:foo]
+    assert_equal :bar, user.notifications.last.params[:foo].to_sym
```
excid3, line 20: This should deserialize as a symbol automatically, so we probably also need the symbol serializer.

test/dummy/config/application.rb
```diff
@@ -8,7 +8,7 @@
 module Dummy
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 6.0
+    config.load_defaults ENV["APPRAISAL"] == "5.2" ? 5.2 : 6.0
```
excid3, line 11: I think we can just delete this in order to always test against the defaults for the currently running Rails version. That should make it simpler. 👍

excid3:
This branch was missing the deserialization step for custom serializers (like Symbols). It would serialize correctly, but not deserialize so I added that functionality in and organized some things.

Check out the branch and see if that works for ya: https://github.com/excid3/noticed/tree/lorint-rails-5-2-compatibility-layer

test/delivery_methods/database_test.rb
lorint: Ah yes, good point.

test/dummy/config/application.rb
lorint: This cleans up a couple of lines, and no more goofing around with the environment variable.

lorint:
> ...
> Check out the branch and see if that works for you

Haha!  Love it!  We were likely doing fairly similar things at the same time on the deserializer side of things.

I love the reorganisation of the polyfill folders.  Can ditch my commit with these similar changes and cherry-pick yours into place if you like, or if you'd like to take it from here then that's fine, too.

Thanks for this!

excid3:
Thanks for merging that! I think this is good to go. Anything else you can think of that's missing?

lorint:
> Anything else you can think of that's missing?

Have renamed serializers.rb to be more generic (since it now loads two of them), and I think we're good.  Have gone through the various tests in our health-related app that uses this, and this works great.

Eager to see Noticed get further developed, and glad to be a part o this.

excid3:
Released as 1.4.0

PR #107, opened 2021-04-06
Update model_generator.rb to be compatible with Rails 5.2

### LINKED ISSUE #106: undefined method 'configs_for' for Rails 5.2
The model_generator.rb throw an error while trying to create a model with Rails 5.2
I suppose because the 'configs_for' method is not present inside Rails 5.2

lorint:
Fixes #106

Why?
  When running "rails g noticed:model" under Rails 5.2, the error
  "NoMethodError: undefined method `configs_for' for #<Hash:...>"
  surfaces.  Rails 6.x and newer has this method, which works
  along with primary and replica database connections.  Rails 5.x
  does not support this behaviour.

How?
  Utilise ActiveRecord::Base#connection_config instead of refering
  to #configurations and from there using #configs_for.  Thankfully
  this is supported by older and newer versions of Rails.

### DIFF
--- a/lib/generators/noticed/model_generator.rb
+++ b/lib/generators/noticed/model_generator.rb
@@ -42,12 +42,20 @@ module Noticed
       end
 
       def params_column
-        case ActiveRecord::Base.configurations.configs_for(spec_name: "primary").config["adapter"]
+        case current_adapter
         when "postgresql"
           "params:jsonb"
         else
           # MySQL and SQLite both support json
           "params:json"
+        end
+      end
+
+      def current_adapter
+        if ActiveRecord::Base.respond_to?(:connection_db_config)
+          ActiveRecord::Base.connection_db_config.adapter
+        else
+          ActiveRecord::Base.connection_config[:adapter]
         end
       end
     end
--- a/test/dummy/config/locales/en.yml
+++ b/test/dummy/config/locales/en.yml
@@ -36,5 +36,4 @@ en:
       message: "This is a notification"
   noticed:
     scoped_i18n_example:
-      message: "This is a custom scoped trnaslation"
-
+      message: "This is a custom scoped translation"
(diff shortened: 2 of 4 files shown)


lib/generators/noticed/model_generator.rb
```diff
@@ -42,7 +42,7 @@ def model_path
       end
 
       def params_column
-        case ActiveRecord::Base.configurations.configs_for(spec_name: "primary").config["adapter"]
+        case ActiveRecord::Base.connection_config[:adapter]
```
excid3, line 45: We may need to use this conditionally because it's being removed:

>DEPRECATION WARNING: connection_config is deprecated and will be removed from Rails 6.2 (Use connection_db_config instead) (called from <main> at (irb):1)

```
ActiveRecord::Base.connection_db_config.adapter
#=> "sqlite3"
```

I'm thinking this would be most reliable:

```ruby
adapter = ActiveRecord::Base.respond_to?(:connection_db_config) ? ActiveRecord::Base.connection_db_config.adapter : ActiveRecord::Base.connection_config[:adapter]
```

lorint: Good call!
Have broken it out into an if / else instead of the ternary because of line length.

Will need the older connection_config[:adapter] version in-place until support ends for Rails 6.x.  (Unless we'd like to also polyfill that one, and I'm not opposed!)

excid3: This will work just fine. 👍

lib/generators/noticed/model_generator.rb
```diff
@@ -42,7 +42,11 @@ def model_path
       end
 
       def params_column
-        case ActiveRecord::Base.configurations.configs_for(spec_name: "primary").config["adapter"]
+        case if ActiveRecord::Base.respond_to?(:connection_db_config)
+               ActiveRecord::Base.connection_db_config.adapter
+             else
+               ActiveRecord::Base.connection_config[:adapter]
+             end
```
excid3, line 49: Can we move this to a method like `def adapter` so we don't have an if inside the case?

lorint: You bet!

PR #121, opened 2021-05-13
Rescuing from database connection errors during the bootstrapping/com…

adriancb:
…pilation phase.

This PR attempts to extend https://github.com/excid3/noticed/commit/0cfcebe4c399fae8b5624235ba4641fc2c4ba153

### DIFF
--- a/lib/noticed/model.rb
+++ b/lib/noticed/model.rb
@@ -1,12 +1,9 @@
 module Noticed
   module Model
     DATABASE_ERROR_CLASS_NAMES = lambda {
-      classes = []
-      if ActiveRecord::VERSION::STRING.to_f > 4.0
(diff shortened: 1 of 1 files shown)


lib/noticed/model.rb
```diff
@@ -1,5 +1,16 @@
 module Noticed
   module Model
+    DATABASE_ERROR_CLASS_NAMES = lambda {
+      classes = []
+      if ActiveRecord::VERSION::STRING.to_f > 4.0
```
excid3, line 5: We only support Rails 5.2+, so I guess I'm wondering if this is needed at all?

coreymartella:
Ahh nice, this will Close #177 as well.

Issue #89, opened 2021-02-05
Batch Deliveries


excid3:
From @SirRawlins

Batching is something I'd love to see.

If we have a notification to deliver to several thousand recipients, then some mail services such a Postmark/Sendgrid/Mailgun allow you to batch send 1,000 messages at a time, converting 000s of API requests into a single call.

Being able to define`DeliveryMethod` classes that were instantiated and enqueued with a collection of recipients, instead of just a single recipient would be nice.

Something like this pseudocode is what I imagine:

```ruby
class DeliveryMethods::Postmark < Noticed::DeliveryMethods::Base

    # Define batch size on the delivery method class.
    deliver_in_batches_of 1_000

    def deliver
        post "https://api.postmarkapp.com/email/send", {
            # Has access to 1_000 recipients, instead of single recipient.
            to: recipients.map(&:email),
            subject: notification.subject
        }
    end

end

class GroupNotification < Noticed::Base

    deliveryby :database
    deliver_by :twilio
    deliver_by :postmark

end

User.all.count
# => 5000

GroupNotification.deliver_later(User.all)
# => Enqueues DeliveryMethods::Database.deliver 5,000 times.
# => Enqueues DeliveryMethods::Twilio.deliver 5,000 times.
# => Enqueues DeliveryMethods::Postmark.deliver 5 times.
```

This would require some refactoring of the `delivery`, `deliver_later` and `run_delivery` methods in `Noticed::Base` to batch and loop recipients if the `DeliveryMethod` supported it.

Not sure if the gains in efficiency and scalability would make it worthwhile though?

_Originally posted by @SirRawlins in https://github.com/excid3/noticed/issues/83#issuecomment-774141149_

excid3:
I think this would make more sense to be a bit more like:

```ruby
delivery_by :postmark # Sends individual emails
delivery_by :postmark, bulk: true # Sends them in bulk
```

```ruby
class DeliveryMethods::Postmark
  def deliver 
  end

  def deliver_in_bulk
    recipients.each_slice(1000) do |group|
      # send group
    end
  end
end
```

That way you could still use Postmark to delivery individually or in bulk.

The question I have is...how does Postmark know when to actually send in bulk? It'd need to queue up the recipients somehow know when it was ready to run.

SirRawlins:
@excid3 yeah that's nice, I like the idea of being able to support both single and batch delivery in a single delivery method. 👍 

[...]

PR #126, opened 2021-05-30
Remove the usage of instance variables from Noticed::Base 

manojmj92:
Since we are attaching `record` and `recipient` to the `Notification` object just before the delivery, there is no longer a requirement to set these from within `Noticed::Base`. 

`recipient` was already being passed as an argument to `run_delivery_method`. Now `record`, if it exists (ie, the database record) is also passed as an argument to the same `run_delivery_method` method.

Both of these will now be available in the `perform` method of `DeliveryMethods::Base`, where it will be attached to the `@notification` object.

I'm trying implement https://github.com/excid3/noticed/issues/89, and this cleanup helps :)

### DIFF
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -9,7 +9,7 @@ module Noticed
     class_attribute :delivery_methods, instance_writer: false, default: []
     class_attribute :param_names, instance_writer: false, default: []
 
-    # Gives notifications access to the record and recipient when formatting for delivery
+    # Gives notifications access to the record and recipient during delivery
     attr_accessor :record, :recipient
 
     delegate :read?, :unread?, to: :record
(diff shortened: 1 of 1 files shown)


manojmj92:
@excid3 Hi Chris, this is ready for review 🙂

lib/noticed/base.rb
```diff
@@ -71,22 +71,19 @@ def params
     def run_delivery(recipient, enqueue: true)
       delivery_methods = self.class.delivery_methods.dup
 
-      # Set recipient to instance var so it is available to Notification class
-      @recipient = recipient
-
       # Run database delivery inline first if it exists so other methods have access to the record
       if (index = delivery_methods.find_index { |m| m[:name] == :database })
         delivery_method = delivery_methods.delete_at(index)
-        @record = run_delivery_method(delivery_method, recipient: recipient, enqueue: false)
```
excid3, line 80: Since we have an `attr_accessor :record` on line 13, assigning to the instance variable makes the record available in all methods later on. We don't need to pass it as an argument since it's an instance variable.

manojmj92: @excid3 But I think this is not being used because we are over-writing the accessor again over [here](https://github.com/excid3/noticed/blob/66d47aa1cc59d4271ec13d525703f5a8ec4b9107/lib/noticed/delivery_methods/base.rb#L40)

"If the accessor's value is being set later, why set it here too?" is the question that led me to this refactor.

manojmj92: Ok, I get it. Its not about the accessor, you'd prefer this to be an instance variable simply because we'd not have to pass `record` around as arguments to methods.

In that case, what do you think about following the same pattern for `recipient` too? The fact that `recipient` is being passed around as argument, while `record` is not is confusing. What do you think about having a similar approach for both - ie, either both remain as instance variables, or both are passed around as arguments? 🙂

excid3: We do actually set the recipient instance variable too: https://github.com/excid3/noticed/blob/66d47aa1cc59d4271ec13d525703f5a8ec4b9107/lib/noticed/base.rb#L75

It's been a long time since I've looked at this code. I don't think we need to pass the recipient then either, or we could pass both. 

I know the main reason we have the attr_accessors for record and recipient on Noticed::Base is so that during a delivery, we can easily access those from the Notification object. As long as we still have that, I think it's fine.

manojmj92: @excid3 personally, I'd prefer passing both as arguments. But happy to do this the other way if you'd prefer that, please let me know 🙂

manojmj92: If there's anything else that needs fixing, let me know. Once this is merged, I can proceed to implement bulk deliveries 🙂

silva96:
Hey I'm a bit too late here but, removing @record on the database notification (former line 80) breaks my code:

```ruby
notification = MyNotification.with(my_param: 'something').deliver(some_user)
if some_user.special_thing
  notification.record.mark_as_read!
end
```

In the new code, calling notification.record is nil

@excid3 @manojmj92

excid3:
#176

silva96:
thanks @excid3 i'll have to wait then until that is merged.

PR #128, opened 2021-06-01
Refactor in preparation for bulk-delivery

manojmj92:
This PR is a refactor in preparation of implementing the bulk-delivery mechanism described in https://github.com/excid3/noticed/issues/89.

I am only doing this across multiple small PRs so that the changes remain small for easy review 🙂 

The crux of this refactor is the fact that:

- We'd need to batch `recipients` for bulk delivery, so looping thru all recipients at the very beginning isn't going to work anymore.
- Hence we pass the whole bunch of `recipients` down, where we will eventually loop thru each recipient.

### DIFF
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -44,20 +44,20 @@ module Noticed
     def deliver(recipients)
       validate!
 
+      recipients = Array.wrap(recipients).uniq
+
       run_callbacks :deliver do
-        Array.wrap(recipients).uniq.each do |recipient|
-          run_delivery(recipient, enqueue: false)
-        end
+        run_all_deliveries(recipients, enqueue: false)
       end
     end
 
     def deliver_later(recipients)
       validate!
 
+      recipients = Array.wrap(recipients).uniq
+
       run_callbacks :deliver do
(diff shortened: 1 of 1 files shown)


lib/noticed/base.rb
```diff
@@ -67,23 +67,31 @@ def params
 
     private
 
-    # Runs all delivery methods for a notification
-    def run_delivery(recipient, enqueue: true)
-      delivery_methods = self.class.delivery_methods.dup
+    def run_all_deliveries(recipients, enqueue: true)
+      individual_delivery_methods = self.class.delivery_methods.dup
```
manojmj92, line 71: This method is where we will eventually be segregating between individual delivery methods and bulk delivery methods based on the `options` provided.

After the segregation, we will eventually have one more method below `run_individual_deliveries`, named `run_bulk_deliveries `

manojmj92:
@excid3 this is ready for review 🙂

excid3:
Could you share the bulk deliveries stuff as a whole? It's hard for me to understand why these changes are helpful without seeing how the whole thing works. 👍

manojmj92:
@excid3 no problem, I will do that 🙂 

Closing this one out.

PR #138, opened 2021-07-27
Parameterless notification helpers

SirRawlins:
I've got a whole bunch of notifications that don't require any additional params, just the recipient and nothing more.

This adds some cheeky little class-level methods to a notification, allowing you to call `NotificationClass.deliver(recipient)`, as a shorthand for `NotificationClass.new.deliver(recipient)` or `NotificationClass.with({}).deliver(recipient)`.

No idea if it's useful, but felt worth sharing.

### DIFF
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -35,7 +35,7 @@ module Noticed
       def deliver(recipients)
         new.deliver(recipients)
       end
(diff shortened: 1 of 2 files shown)


excid3:
Makes sense to me. Thanks @SirRawlins 👍

PR #142, opened 2021-07-30
Add stream customization for ActionCable delivery method

excid3:


excid3:
Does this work for you @jose2007kj?

jose2007kj:
apologies @excid3 for the delay in response,i didn't use this method as i didn't know how to install this or update my gem-file locally [all i know is bundle install]  

since recipient is unique key i just decided to use it[i didn't know that it was unique when i created the issue, and i realized it when u explained it]

I am sure this feature will be useful, when user want custom stream name

sorry if this caused u any inconvenience 
thanks

PR #143, opened 2021-08-04
fix: make recipient available in notifcation after `to_notification`

RolandStuder:
## Problem

I wanted to test a notification message, from the database, using `Notification.find(x).to_notification`, but even though the recipient was set on the record it was not available. As our `message` relies on aspects from the recipient, calling `message` raises an error on `recipient.some_thing`, as recipient is nil.

## Resolution

Set recipient on `to_notification`

## Workaround

One can of course just set the recipient directly, as you would when making a unit_test. But I would expect to have `recipient` to be set, when using `to_notification`.

Let me know, if you would like some tests, and what they should cover.


excid3:
Good catch @RolandStuder 👍 

It would be good to add a test for this. I think if you make an example notification with a message that references recipient, that would cover it.

Something like:

```ruby
class RecipientExample
  deliver_by :database
  
  def message
    recipient.id
  end
end

test "sets recipient when loading notification from database" do
  RecipientExample.deliver(@user)
  assert_equal @user.id, Notification.last.to_notification.message
end
```

PR #150, opened 2021-09-17
Allow Noticed to work with PostGIS ActiveRecord adapter

csutter:
We use [activerecord-postgis-adapter][1] in one of our applications, which for all intents and purposes is a superset of the default ActiveRecord `postgres` adapter.

Because Noticed explicitly checks whether the ActiveRecord adapter name matches `postgres`, our automatic deletion of notifications for deleted records using `has_noticed_notifications` no longer works.

- Add `postgis` as an equal alternative to `postgres` when checking for adapters in `has_noticed_notifications`
- Fix a tiny spelling mistake

[1]: https://github.com/rgeo/activerecord-postgis-adapter

### DIFF
--- a/lib/noticed/has_notifications.rb
+++ b/lib/noticed/has_notifications.rb
@@ -1,6 +1,6 @@
 module Noticed
   module HasNotifications
-    # Defines a method for the association and a before_destory callback to remove notifications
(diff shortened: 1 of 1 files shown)


cesidio (approved):

excid3:
Looks good to me, thanks @csutter 👍

PR #161, opened 2021-10-08
Allow dynamic mailer to be returned if symbol passed

excid3:
This allows the `mailer: :custom_mailer` option to be set to several different options now:

1. String - constantized to class
2. Symbol - calls method that should return mailer class
3. Class - Referencing the class directly

@tbcooney give this a try?

### DIFF
--- a/lib/noticed/delivery_methods/email.rb
+++ b/lib/noticed/delivery_methods/email.rb
@@ -9,6 +9,9 @@ module Noticed
 
       private
 
+      # mailer: "UserMailer"
+      # mailer: UserMailer
+      # mailer: :my_method - `my_method` should return Class
       def mailer
         option = options.fetch(:mailer)
         case option
@@ -21,8 +24,18 @@ module Noticed
         end
       end
 
+      # Method should be a symbol
+      #
+      # If notification responds to symbol, call that method and use return value
+      # If notification does not respond to symbol, use the symbol for the mailer method
+      # Otherwise, use the underscored notification class name as the mailer method
       def method
-        options[:method] || notification.class.name.underscore
+        method_name = options[:method]&.to_sym
+        if method_name.present?
+          notification.respond_to?(method_name) ? notification.send(method_name) : method_name
+        else
+          notification.class.name.underscore
+        end
       end
 
       def format


tbcooney:
There are a few quirks with the way `method` works in the delivery methods....in short, it would be great to set the method to use different mailers, but it's called form the base Noticed class so you don't have access to `record` or `params`

```
deliver_by :email, mailer: "NotificationMailer", method: :#{params[:note].noteable_type.underscore}_mention
#=> This does not work

deliver_by :email, mailer: "NotificationMailer", method: :mention_notification
#=> This works
```

excid3:
@tbcooney Had a chance to try this again?

ausangshukla:
How do I do this in the new V2.1? 

deliver_by :email do |config|
      config.mailer = :mailer_name
      config.method = :email_method
      config.params = :email_data
end

The :mailer_name is a method, but its never called and the gem tries to use a mailer called :mailer_name and fails.

excid3:
@ausangshukla doesn't do dynamic [lookup for constants](https://github.com/excid3/noticed/blob/main/lib/noticed/delivery_method.rb#L42), but we can add that.

ausangshukla:
Changing the mailer option to use evaluate_option as below works. This was possible in the version < 2, hence need this change. Thanks

`module Noticed
  module DeliveryMethods
    class Email < DeliveryMethod
      required_options :mailer, :method

      def deliver
        mailer = evaluate_option(:mailer)
        email = evaluate_option(:method)
        args = evaluate_option(:args) || []
        mail = mailer.with(params).send(email, *args)
        (!!evaluate_option(:enqueue)) ? mail.deliver_later : mail.deliver_now
      end

      def params
        (evaluate_option(:params) || notification&.params || {}).merge(
          notification: notification,
          record: notification&.record,
          recipient: notification&.recipient
        )
      end
    end
  end
end
`

excid3:
The `fetch_constant` is special because we want to handle String specially for constant lookups, but non-strings can use `evaluate_option` so it handles it the same otherwise. Will be a nice improvement. Thanks for catching this @ausangshukla . 👍

PR #164, opened 2021-10-15
Fix postgis support for model generator

bmorrall:
Attempted to fix postgis support to the model adaptor, but Christian Sutter beat me to it.

The model generator is still broken, so added a fix for it here.

When's the next version of the gem coming out?


excid3:
Hopefully soon. I was hoping to get the bulk database delivery going, but we'll see if that makes it in or not.

PR #173, opened 2021-11-05
feat: add enqueue option to email

coder2000:
Adds an option to email delivery that uses ActiveJob queues.


excid3:
Just modified the test to make sure it enqueued an email and then documented that `enqueue` is false by default.

Thanks for the quick work @coder2000 🙏

stephaneliu:
For my edification, why would someone want to enqueue email delivery when the caller is also executed from an enqueued process?

excid3:
Might be handled by different workers.

PR #175, opened 2021-11-09
Fix invalid notification token cleanup

### LINKED ISSUE #174: Invalid iOS token can't be cleaned up
[This line](https://github.com/excid3/noticed/blob/5f8afbdab0ece3b6b08004a59a883662405214ee/lib/noticed/delivery_methods/ios.rb#L21) raises an exception if the response is invalid when sending an iOS push notification. If the token is invalid (e.g. the app was uninstalled off of a device) the response returns 410 or 400, which will return false for `response.ok?`. So the following line, where the cleanup would occur, will never get called.

I'm planning on fixing this issue but wanted to let folks know in case they run into it themselves.

joemasilotti:
Fixes #174 by...

1. Only raising an error if the token was valid _and_ the push failed
2. Fixing a typo in the callback method

### DIFF
--- a/lib/noticed/delivery_methods/ios.rb
+++ b/lib/noticed/delivery_methods/ios.rb
@@ -18,10 +18,13 @@ module Noticed
 
             response = connection.push(apn)
             raise "Timeout sending iOS push notification" unless response
-            raise "Request failed #{response.body}" unless response.ok?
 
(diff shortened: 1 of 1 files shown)


excid3:
Looks good. Thanks @joemasilotti 🙏

PR #176, opened 2021-11-12
Explicitly assign record as instance variable

leighhalliday:
I believe that it is assigning record as a local variable rather than to the instance variable. By adding `self.` it should call the attr_accessor method so that it is available later via `notification.record`.

I have added a test to look for this which went from red to green as I changed how the record variable was being assigned. (We rely on this to be here in some backfill functionality so the 1.5 release caused some tests to fail).


excid3:
Doh, I totally missed that in the previous PR. Thanks for fixing that and adding the test. 🙏

excid3:
Released as 1.5.1 👍

PR #180, opened 2021-11-18
Set recipient instance variable in Noticed::Base

### LINKED ISSUE #179: `nil` recipient in email callback after updating to 1.5+
We've been getting an issue where the recipient is nil in the `after_email` callback of a notification. We use `deliver_later`, which I realize gets called around the enqueuing of the delivery (and not explicitly after). The recipient is nil only during the `after_email` callback, but is present in both the `database` and `email` delivery methods.

My guess is that the removal of the `@recipient` instance variable in #126 here caused an issue but haven't had a chance to fully debug yet: https://github.com/excid3/noticed/pull/126/files#diff-8d74699b16f13d2f080f755ae12faee3a7cd24dd19833cf203f97bb7278db400L74-L76

We've seen this issue with 1.5.0, 1.5.1, and 1.5.2, but downgrading to 1.4.1 fixed it for us.

The rough setup of our notification classes:

```rb
class MessageNotifier
  include Sidekiq::Worker

  def perform(message_type, message_url)
    message = MessageService.find_message_by_url(message_type, message_url)
    MessageService.unreads(message_url).each do |user_id, unread_ts|
      recipient = User.find(user_id)
      next if recipient.nil?

      params = { unread_ts: unread_ts, message: message }
      UnreadMessagesNotification.with(params).deliver_later(recipient) # This is the notification call
    end
  end
end

class UnreadMessagesNotification < Noticed::Base
  deliver_by :database
  deliver_by :email, mailer: 'MessageMailer',
                     method: :unread_messages_email,
                     format: :generate_mailer_params

  after_email :mark_recipient_as_read
 
  param :unread_ts, :message

  def mark_recipient_as_read
    user_id = recipient.id # Recipient is nil here
    MessageService.mark_as_read(user_id, message, as_of: unread_ts)
  end

  def generate_mailer_params
    {
      unread_ts: unread_ts,
      message: message
    }
  end
end
```

clinejj:
Fixes #179 

This is a partial revert of #126 using style from #176.

In #179, issue was identified where the `recipient` was nil in `before` callbacks on delivery methods. This fixes that issue and adds appropriate tests.

_______
There were some other issues I identified while trying to increase test case coverage for callback methods, namely:

- Adding the following callback to `CallbackExample` in `noticed_test` would raise the argument error that `recipient` is nil, it is unclear if that is intended behavior:

```rb
before_deliver do
  raise ArgumentError unless recipient

  self.class.callbacks << :before_everything
end
```

- Adding an `around_deliver` callback to Noticed::DeliveryMethods::Test would fail multiple tests, as the `#deliver` method would not get called

- Adding before/around/after callbacks to CallbackExample on both everything and the database method would only call the everything callbacks, and not the method callbacks. For example, adding `before_deliver` and `around_deliver` callbacks to the example test would result in the test returning `[:before_everything, :around_everything, :after_everything]`

There's some unclear logic as to what callbacks get called when, and I'm not sure what is expected vs what happens in test vs what happens in an actual implementation, but happy to log to another issue or document elsewhere.

### DIFF
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -81,7 +81,7 @@ module Noticed
     def run_delivery(recipient, enqueue: true)
       delivery_methods = self.class.delivery_methods.dup
 
-      @recipient = recipient
+      self.recipient = recipient
 
       # Run database delivery inline first if it exists so other methods have access to the record
       if (index = delivery_methods.find_index { |m| m[:name] == :database })
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -80,7 +80,7 @@ module Noticed
     # Runs all delivery methods for a notification
     def run_delivery(recipient, enqueue: true)
       delivery_methods = self.class.delivery_methods.dup
-      
+
       @recipient = recipient
 
       # Run database delivery inline first if it exists so other methods have access to the record
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -80,6 +80,8 @@ module Noticed
     # Runs all delivery methods for a notification
     def run_delivery(recipient, enqueue: true)
       delivery_methods = self.class.delivery_methods.dup
+      
+      @recipient = recipient
 
       # Run database delivery inline first if it exists so other methods have access to the record
       if (index = delivery_methods.find_index { |m| m[:name] == :database })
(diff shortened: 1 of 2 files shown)


excid3:
Thanks @clinejj! Good to have some more robust tests here for next time internals get refactored. 😅 

The order of callbacks kind of vary.

We'll always run the database delivery first. Depending if the deliveries are async or not, the delivery methods will either run in parallel or in order.

Issue #182, opened 2021-11-24
iOS delivery method does not allow cert_path to be a string


erbunao:
I encountered an issue when a custom `cert_path` method  returns a `StringIO.new("certificate_content_here")` object instead of a file.

If I have this code below and an empty `config/certs/ios/apns.p8` file, the push notification gets delivered without an issue.

```
class SomeNotification
  deliver_by :ios, {
    format:            :ios_format,
    cert_path:         ios_cert_path,
  }
  
  def ios_cert_path
    StringIO.new(Rails.application.credentials.dig(:ios, :apn_certificate).to_s)
  end
end

```

However if the p8 file gets deleted, this exception shows up

```
ArgumentError: Could not find APN cert at '.../config/certs/ios/apns.p8'
from .../noticed-1.5.2/lib/noticed/delivery_methods/ios.rb:12:in `deliver'
```

I believe the code should be able to handle a `cert_path` with a string content and not restrict it from always getting the certificate from a file.

excid3:
StringIO isn't currently supported, but I'd love to! Just hadn't thought about that yet. 😜

Your option also needs to be a symbol so that it calls the method with the matching name (just like an ActiveRecord callback). So you should have:

```ruby
  deliver_by :ios, {
    format:            :ios_format,
    cert_path:         :ios_cert_path,
  }

  def ios_cert_path
  end
```

erbunao:
Ah thanks for clarifying that @excid3!

I wonder how would you manage getting the right cert_path depending on the environment (production vs staging vs development)

excid3:
If you had 
```
config/certs/ios/production.p8
```

erbunao:
Yeah totally! I'm worried to add in the p8 files in the repo as I believe they should be kept in a private place. Therefore I was using Rails credentials to get the content.

Has this file management ever been a concern before?

joemasilotti:
FWIW, I usually commit my p8 files to the repo and put the team ID and key in Rails credentials. That way even if someone got the p8 they couldn't use it unless they also got access to your credentials.

erbunao:
@joemasilotti that makes complete sense! 

Just looping back on the main point, I feel the `cert_path` interface would benefit from having a parity on `Apnotic::Connection`'s interface as well, like so

```
Apnotic::Connection.new(cert_path: StringIO.new("pem cert as string"))
```

Would that be something beneficial from you perspective as well @excid3 @joemasilotti ?

excid3:
[...]

PR #183, opened 2021-11-24
Support for StringIO input type for cert_path option

erbunao:
This PR adds the flexibility for the `cert_path` iOS delivery method to receive a `StringIO` object type similar to `Apnotic::Connection`'s like so

```
Apnotic::Connection.new(cert_path: StringIO.new("cert as string"))
```

Relevant Issue: https://github.com/excid3/noticed/issues/182

### DIFF
--- a/lib/noticed/delivery_methods/ios.rb
+++ b/lib/noticed/delivery_methods/ios.rb
@@ -154,8 +154,8 @@ module Noticed
 
       def valid_cert_path?
         case cert_path
-        when StringIO
-          cert_path.read.present?
+        when File, StringIO
+          cert_path.size > 0
         else
           File.exist?(cert_path)
         end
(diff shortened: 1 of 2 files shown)


lib/noticed/delivery_methods/ios.rb
```diff
@@ -152,6 +152,15 @@ def development?
         end
       end
 
+      def valid_cert_path?
+        case cert_path
+        when StringIO
```
excid3, line 157: Let's also add `File` here too, just in case the cert has already been read for whatever reason.

```suggestion
        when File, StringIO
```

excid3:
Could you also add examples to the iOS delivery method doc? 

https://github.com/excid3/noticed/blob/master/docs/delivery_methods/ios.md

lib/noticed/delivery_methods/ios.rb
erbunao: Great idea!

erbunao:
@excid3 Done updating the docs and to support `File` object type as well 😄

erichmachado:
This looks great! I wonder if it could be included in the next release, @excid3?

excid3 (approved):

PR #190, opened 2022-01-06
Fcm

cjilbert504:
This PR is the initial go at setting up Android notifications via FCM.


excid3:
So excited about this! 🎉 

I think these are the outstanding things now that we have a working HTTP request.

- [ ] Add delivery method options for credentials (and project ID if we decide to keep that)
- [ ] Add option for customizing format
- [ ] Decide if we want to refactor and extract the project ID from the credentials json
- [ ] Tests

excid3:
Thanks for all the great work on this @cjilbert504!

Issue #169, opened 2021-10-27
Replace http with Faraday


excid3:
Using Faraday for HTTP requests provides a lot more flexibility as the user can define which HTTP client is used.

- [ ] Remove HTTP gem
- [ ] Add Faraday
- [ ] Replace HTTP requests with Faraday
- [ ] Configurable Faraday client (See Vultr.rb for an example: https://github.com/tolbkni/vultr.rb/blob/master/lib/vultr/client.rb#L99-L109)

excid3:
We're going to roll with net/http so we don't need any additional dependencies.

PR #199, opened 2022-01-23
replace http gem with faraday

yarafan:
Closes https://github.com/excid3/noticed/issues/169.

Brief summary:

* HTTP gem is replaced with Faraday (using standard ruby net_http as adapter)
* `Notice::DeliveryMethods::Base#post` method was rewritten to use Faraday
* Provide possibility to create custom Faraday connection via #build_connection in custom delivery methods

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     noticed (1.5.7)
-      http (>= 4.0.0)
+      faraday (>= 0.17.3, < 2.0)
       rails (>= 5.2.0)
 
 GEM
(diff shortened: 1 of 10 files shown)


Gemfile.lock
```diff
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     noticed (1.5.7)
-      http (>= 4.0.0)
+      faraday (>= 0.17.3, < 2.0)
```
yarafan, line 5: To provide compatibility with `googleauth (1.1.0)` which also depends on `faraday (>= 0.17.3, < 2.0)`
I suppose faraday can be bumped to 2.1 with googleauth but I think it  should be done in separate request

excid3:
Rather than introducing another dependency, we're going to use net/http for simplicity.

PR #202, opened 2022-01-27
Handling FCM error

### LINKED ISSUE #200: Handling FCM error
When trying to push notification unsuccessfully using `FCM`, I got this error 
```ruby
NoMethodError (undefined method `code' for #<Noticed::ResponseUnsuccessful: Noticed::ResponseUnsuccessful>)
```
while debugging this I found that errors returns in this format 

```json
{
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Unknown name \"token\" at 'message': Proto field is not repeating, cannot start list.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.BadRequest",
        "fieldViolations": [
          {
            "field": "message",
            "description": "Invalid JSON payload received. Unknown name \"token\" at 'message': Proto field is not repeating, cannot start list."
          }
        ]
      }
    ]
  }
}
``` 

it seems that the `code` is a child of the `error` object, we can not access it direct form the response

HeshamMagdy97:
closes #200

### DIFF
--- a/lib/noticed/delivery_methods/fcm.rb
+++ b/lib/noticed/delivery_methods/fcm.rb
@@ -20,7 +20,7 @@ module Noticed
         device_tokens.each do |device_token|
           post("#{BASE_URI}#{project_id}/messages:send", headers: {authorization: "Bearer #{access_token}"}, json: {message: format(device_token)})
         rescue ResponseUnsuccessful => exception
-          if exception.code == 404
+          if exception.response.parse.dig('error', 'code') == 404
             cleanup_invalid_token(device_token)
           else
             raise
--- a/lib/noticed/delivery_methods/fcm.rb
+++ b/lib/noticed/delivery_methods/fcm.rb
@@ -20,7 +20,7 @@ module Noticed
         device_tokens.each do |device_token|
           post("#{BASE_URI}#{project_id}/messages:send", headers: {authorization: "Bearer #{access_token}"}, json: {message: format(device_token)})
         rescue ResponseUnsuccessful => exception
-          if exception.response.parse.dig('error', 'code') == 404
+          if exception.response.code == 404
             cleanup_invalid_token(device_token)
           else
             raise


cjilbert504 (approved):
Thank you again @HeshamMagdy97 for the submission. This should do the job nicely :)

lib/noticed/delivery_methods/fcm.rb
```diff
@@ -20,7 +20,7 @@ def deliver
         device_tokens.each do |device_token|
           post("#{BASE_URI}#{project_id}/messages:send", headers: {authorization: "Bearer #{access_token}"}, json: {message: format(device_token)})
         rescue ResponseUnsuccessful => exception
-          if exception.code == 404
+          if exception.response.parse.dig('error', 'code') == 404
```
excid3, line 23: We should be able to use `exception.response.code` here instead of having to pull the code out of the JSON. I'm going to make an edit to the PR and see if that works.

HeshamMagdy97: Hi @excid3 Thanks for your feedback, but I think google always return `200` as status code 
see this trace code returned from `fcm`
```ruby 
[2022-02-03T15:16:47.924301 #90775] DEBUG -- : [httplog] {"method":"POST","url":"http://www.googleapis.com:443/oauth2/v4/token","response_code":200,"benchmark":0.1543269995599985}
400 Bad Request
{
  "error": {
    "code": 400,
    "message": "The registration token is not a valid FCM registration token",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.firebase.fcm.v1.FcmError",
        "errorCode": "INVALID_ARGUMENT"
      }
    ]
  }
}
```

so I think we must check the `code` entry inside `error` object.

excid3: That's the info I needed. :+1:

Can you update and include a comment in the code?

excid3: The logs show `400 Bad Request` so it looks like the status code is `400`, not `200`?

I think we'll need to check if the code is `400` or `404` then.

nicalpi:
Just letting you know that we got caught up on that one today. The solution, until version 1.5.8 is released is to use master as this code is not present on the 1.5.7 (if anyone else is passing here with the issue)

excid3:
I'll release that right now.

PR #214, opened 2022-03-12
Demodulize class name on mailer method lookup

### LINKED ISSUE #208: Demodulize class name on mailer method lookup
Hi, mailer method is defined as an underscored class name if the `method` option wasn't provided

https://github.com/excid3/noticed/blob/df561f348cc33283e4cc62f711f7d9a47b811a12/lib/noticed/delivery_methods/email.rb#L36-L43

But if I put my notification class to a module the result string will contain the name of that module as a prefix.

```ruby
Employees::NotifyAboutNewOrder.name.underscore #=> "employees/notify_about_new_order"
```

I think it is appropriate to pre-demodulize a class name here.

StanisLove:
Resolves #208

### DIFF
--- a/lib/noticed/delivery_methods/email.rb
+++ b/lib/noticed/delivery_methods/email.rb
@@ -38,7 +38,7 @@ module Noticed
         if method_name.present?
           notification.respond_to?(method_name) ? notification.send(method_name) : method_name
         else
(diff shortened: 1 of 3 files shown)


aka-nez:
I just ran into the same issue. PR make sense for me

PR #215, opened 2022-03-16
Refactor comment

liroyleshed:
Extract logic into a private method.

### DIFF
--- a/lib/noticed.rb
+++ b/lib/noticed.rb
@@ -30,8 +30,7 @@ module Noticed
       notification.notify(recipient)
     end
 
-    # Clear the recipient after sending to the group
-    notification.recipient = nil
+    clear_recipient(notification)
   end
 
   mattr_accessor :parent_class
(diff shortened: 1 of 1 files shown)


lib/noticed.rb
```diff
@@ -30,8 +30,7 @@ def self.notify(recipients:, notification:)
       notification.notify(recipient)
     end
 
-    # Clear the recipient after sending to the group
-    notification.recipient = nil
+    clear_recipient(notification)
```
excid3, line 33: What about adding the method to the Notification instead? 

```suggestion
notification.clear_recipient
```

liroyleshed: Sounds great. Let's do a public instance method on the Base class?

excid3: I think that makes sense. 👍

liroyleshed: I think it's good to go. 👍

PR #222, opened 2022-04-24
Added Slack Message delivery method

igor-alexandrov:
Hi Chris,

This PR adds ability to deliver notifications via Slack Messages API. Internally it uses [chat.postMessage](https://api.slack.com/methods/chat.postMessage) call.

It is still WIP and should be polished, but as of today it works and we use it in our projects.


excid3:
Awesome @igor-alexandrov. It's been a while since I used Slack. How does this differ from the webhook Slack integration we currently have? We'll need to make sure the difference is explained in the docs.

Also, how hard would it be to drop the slack client dependency? I've been wanting to change everything to using net-http so we can drop the need for Faraday/an http gem and just rely on Ruby. Still need to update the existing ones, but it's on my todo list.

igor-alexandrov:
@excid3 thanks for your response!

Using `chat.postMessage` API call allows you to interact with Slack by tokens obtained from OAuth2 flow. In this case you can message to users, channels or apps. While webhooks only allow to message to channel where webhook is installed. You can read more here: https://api.slack.com/messaging/webhooks and here is discussion on StackOverflow: https://stackoverflow.com/questions/55896335/slack-chat-postmessage-vs-incoming-webhook.

Getting rid of `faraday` is great idea! I will try to do it.

igor-alexandrov:
@excid3 I removed `slack-ruby-client` dependency and switched to `Net::HTTP`. Also updated documentation.

I also consider also renaming `slack` delivery method to `slack_webhook` and mark `slack` as deprecated. What do you think about this?

igor-alexandrov:
@excid3 Chris, I also think it would be great to have stubs inside of delivery methods.
Here is an example of how this can be implemented in `SlackMessage`:

```ruby
module Noticed
  module DeliveryMethods
    class SlackMessage < Base
      URL = "https://slack.com/api/chat.postMessage"

      def self.stub_success_response(&block)
        response = MiniTest::Mock.new
        response.expect :body, '{"ok": true}'

        SlackMessage.stub_any_instance :post, response do
          yield block
        end
      end
     
      #...
  end
end
```

This requires additional dependency of https://github.com/codeodor/minitest-stub_any_instance, but then you can test your notifications easier:

```ruby
context 'when there is a slack_authorization' do
  let(:member) { build_stubbed(:member, slack_authorization: slack_authorization) }

  it 'delivers Slack Message notification' do
    allow(block_double).to receive(:render)

    expect(Slack::Blocks::Contract::NewContractBlock).to receive(:new).with(contract: contract).once.and_return(block_double)

    Noticed::DeliveryMethods::SlackMessage.stub_success_response do
      notification.deliver(member)
    end
  end
end
```

Tell me, what do you think about this.

igor-alexandrov:
@excid3 Chris did you have a chance to think on this?

igor-alexandrov:
@excid3 ping ping.

samtgarson:
A big issue we've encountered when trying to implement this gem—the Incoming Webhook method of sending messages with the Slack API only allows posting to the single channel which the user chose when they authenticated.

Our use case (and I imagine quite a common one) is to send notifications which aren't totally public in their scope—so we need to send them as private/direct messages, which requires something shaped more like this PR.

@excid3 this would be extremely useful for us, we're looking at borrowing this code/targeting this branch. Any chance of getting this merged?

excid3:
Yeah I'm good with this. It will be a breaking change though, so we'll have to cut this as a 2.0 release probably.

samtgarson:
Amazing, thanks!

If we want to avoid a 2.0 release, we could keep the original Slack delivery method named as it is? Not sure if that would make the release process easier. 

Looks like the tests failed on the Setup Ruby step anyway 😳

excid3:
If we keep the Slack delivery method the same, then we wouldn't have to make a 2.0 release. 👍 

The setup ruby issue should be able to be resolved by merging in master I think. I wasn't able to do it properly on github.com and don't have time to do it manually right now.

samtgarson:
@igor-alexandrov do you have any time to get this over the line? Otherwise I'd be happy to fork and give it a final push 👍

igor-alexandrov:
Guys, I rebased my PR on `master`. For now, `:slack` delivery method still works and I deprecated it. If you think I should revert `:slack_webhook` back to `:slack`, I will do it.

igor-alexandrov:
@excid3 can you please take a look?

multiplegeorges:
This looks great, @igor-alexandrov! We've been using some homegrown code for this as well and it would be great to consolidate all our notifications into noticed.

I know it's not merged yet, but I have one request:

On a synchronous notification with`#deliver`, would it be possible to return the message ID from Slack so that subsequent notifications to the same message ID would update the message in Slack instead of posting a new one?

The API for updating a message is [chat.update](https://api.slack.com/methods/chat.update).

Let me know what you think, this may be out of scope for noticed and this PR.

igor-alexandrov:
@multiplegeorges thanks a lot for your feedback.

Actually I don't think that it is a good idea to return message ID as a result of `#deliver`. It would be better to come to some unified result for all delivery methods, but this is something we can think about in 2.0 release.

For now there is a undocumented 😈 method `#on_success` https://github.com/igor-alexandrov/noticed/blob/master/lib/noticed/delivery_methods/slack_message.rb#L30 that you can define in your `Notification`, it will receive response from Slack. Below is an example.

``` ruby
# frozen_string_literal: true

class TimeOff::CreatedNotification < Noticed::Base
  deliver_by :slack_message,
    format: :format_slack_message,
    on_success: :on_success_slack_message,
  
  def format_slack_message
  # ...
  end

  def on_success_slack_message(result)
    data = result.slice('ts', 'channel')
    params[:approve_request].update(slack_message_data: data) if data.any?
  end
end

```

multiplegeorges:
Agreed, a general solution would be great and would likely be a breaking change necessitating 2.0 release.

I'll have a look at `on_success`, thanks 😈

igor-alexandrov:
Rebased PR on master. Is there a way to make this finally merged?

excid3:
Slowly but surely getting around to this. 👍

excid3:
Slowly but surely getting around to this. 👍

PR #223, opened 2022-04-27
added ActionMailer::Preview documentation to email docs

MichaelLambs:
Added some documentation on how to create an `ActionMailer::Preview` as this took me a bit to figure out, however it may be obvious to others.

- Tricky part for me was that when using the noticed gem I had to get the `recipient` param plus a few others  into the `mailer_method`.
- There may be a better spot for documenting this.


excid3:
Thanks @MichaelLambs 👍

PR #233, opened 2022-06-06
Remove unused method

cjilbert504:
I _think_ we are able to drop this method. If this method were to be called, I believe it would error out as there no longer is a `notify` method on notifications. The `notify` method was renamed to `deliver` in this [PR](https://github.com/excid3/noticed/commit/73186a2da05b37e3babe30afbb89361a769980b7#diff-8d74699b16f13d2f080f755ae12faee3a7cd24dd19833cf203f97bb7278db400L25) and I believe that this method is a lingering artifact from before the change occurred. 

Also moved the `Fcm` autoload line into correct alphabetical position :)


excid3:
I think this is safe to remove too. I don't believe we ever documented it and it's cleaner to send a notification directly from the class.

Issue #213, opened 2022-03-11
Add generic Webhook delivery method


excid3:
We should add a generic webhook delivery method.

Discord, Slack, etc just send a webhook, so possibly they can be refactored so that they inherit from the webhook delivery method.

PR #255, opened 2022-10-18
Slack delivery method inherits Webhook

hmdros:
Hi Chris,

Me and @toppa are making contributions on `hacktoberfest`.

Our PR is based on @vickyonit 's previous suggestion on the issue: https://github.com/excid3/noticed/issues/213

We had the Slack delivery method inherit the Webhook class so we could reuse some methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Currently, we support these notification delivery methods out of the box:
 * Vonage / Nexmo (SMS)
 * iOS Apple Push Notifications
 * Firebase Cloud Messaging (Android and more)
+* Webhook
 
 And you can easily add new notification types for any other delivery methods.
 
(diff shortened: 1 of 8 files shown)


toppa:
Hi @excid3 - we'd be happy to do more on this. We started by trying to refactor the Webhook class from @vickyonit 's PR, to match the convention of the project (i.e. we started with a `webhook` subdir containing a `base.rb` class, etc). But this approach interfered with the abstraction logic for determining the delivery method, which won't work with that approach (at least not without some redesign).

Also you could treat the Webhook class more like an interface, by having all the delivery methods inherit from it, but that might be getting too Java-y. Anyway, let us know what you think and we'd be happy to proceed based on your feedback.

PR #258, opened 2022-11-25
test: add notification object generator test

nachiket87:
adds a generator test for the notification object generator.

`rails generate noticed:notification CommentNotification`

https://github.com/excid3/noticed#-usage


nachiket87:
Hey @excid3 sorry to tag you here but I think my PR will go un-"noticed" 😅.

excid3:
I have a newborn baby at home, so I've not had a lot of free time lately. We will get to this. 👍 

One of the things we plan to do for the next version is rename the Notification to Notifier and leave Notification as the name for the database model. That way it's less confusing.

Maybe we should start here by renaming it to `noticed:notifier` and we can change the folder to `app/notifiers` later

PR #261, opened 2022-12-06
Added initial prototype for Noticed Web UI

igor-alexandrov:
This PR adds initial prototype for Noticed Web UI.

What is included:
- Sinatra App with Tailwind CSS UI 
- Utility classes and modules (helpers, paginator)
- Listing of Database notifications

What I currently thinking about?
- What additional data can we show?
- What to put on dashboard tab?

My suggestions:
- Add statistics (sent, failed per each notification, delivery channel)

Screenshot 2022-12-06 at 00 41 03


igor-alexandrov:
@excid3 do you have any ideas about features that can be added to the UI?

excid3:
Would be cool to have this as a separate gem now that v2 improves the models. Don't think we'll keep this in the gem itself to keep things simple and easier to maintain.

PR #279, opened 2023-04-04
Avoid override of Object#method

quadule:
`Noticed::DeliveryMethods::Email` overrides `Object#method`, which can cause it to behave unexpectedly. I was trying to use `rescue_from` in the parent job class, but it [blows up here with a `NoMethodError`](https://github.com/rails/rails/blob/7c70791470fc517deb7c640bead9f1b47efb5539/activesupport/lib/active_support/rescuable.rb#L103-L106) because the overridden `method` is private.


excid3:
Thank you @quadule! Surprised we missed that until now! Great catch. 👍

PR #283, opened 2023-05-05
feat: add ability to have dynamic queue names

armiiller:
Allow queue names to be a symbol (for example you might want to queue notifications differently based on subscription level).

### DIFF
--- a/lib/noticed/base.rb
+++ b/lib/noticed/base.rb
@@ -113,7 +113,12 @@ module Noticed
 
         # If the queue is `nil`, ActiveJob will use a default queue name.
         queue = delivery_method.dig(:options, :queue)
-        queue = send(queue) if queue.is_a? Symbol
+
+        # if the queue is a proc or lambda, call it with the notification instance
+        if queue.respond_to?(:call)
+          # set the local queue variable and the option in args we are passing to the job
+          args[:options][:queue] = queue = queue.call(self)
+        end
 
         # Always perfrom later if a delay is present
         if (delay = delivery_method.dig(:options, :delay))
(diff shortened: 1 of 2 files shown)


lib/noticed/base.rb
```diff
@@ -113,6 +113,7 @@ def run_delivery_method(delivery_method, recipient:, enqueue:, record:)
 
         # If the queue is `nil`, ActiveJob will use a default queue name.
         queue = delivery_method.dig(:options, :queue)
+        queue = send(queue) if queue.is_a? Symbol
```
excid3, line 116: If users are already passing in the queue as a Symbol, this would be a breaking change. I think that's pretty common with ActiveJob because their docs show `queue_as :feeds`.

A better solution might be to accept a lambda or Proc and check if queue responds to `call`?

armiiller: I would have to test again, but I think this portion of noticed currently breaks if you pass in a symbol.

However, yes, I agree, with ActiveJob using symbols this might better be checked as you said, with `respond_to?(:call)`.

Let me make the changes and do some testing 👍

excid3: Sounds like we should add some test cases too then. 😜

armiiller: Sounds good, i will bake tests in too.

Give me a few weeks on this one. Have family stuff next two weeks, so my time might be sparse when I am at the computer

excid3: No worries! I might work on this myself in the meantime.

armiiller: Made the talked about changes. I also added tests both supporting that procs work and symbol names stay as strings 👍

excid3: Excellent!

armiiller: Just pushed the typo error

excid3:
This should be supported in v2. 👍

PR #284, opened 2023-05-11
Enable the :debug and :ignore_failure options

mike-burns:
In c78eab6145f5a4cef36e6628073ec9cc61258bb3, the `options` accessor was shadowed by a local variable set to the empty hash. This prevented `#post` from using the `:debug` and `:ignore_failure` options.

This removes the shadowing, allowing those options to work again.

### DIFF
--- a/test/delivery_methods/slack_test.rb
+++ b/test/delivery_methods/slack_test.rb
@@ -34,4 +34,22 @@ class SlackTest < ActiveSupport::TestCase
 
     assert_kind_of HTTP::Response, response
   end
+
+  test "logs verbosely in debug mode" do
+    class ActiveSupport::Logger
+      attr_reader :logging_log
+      def debug(msg)
+        @logging_log ||= []
+        @logging_log << msg
+      end
+    end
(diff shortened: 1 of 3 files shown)


test/delivery_methods/slack_test.rb
```diff
@@ -34,4 +34,22 @@ def slack_url
 
     assert_kind_of HTTP::Response, response
   end
+
+  test "logs verbosely in debug mode" do
+    class ActiveSupport::Logger
```
mike-burns, line 39: I confess, I don't know the best way to stub in this test framework. Interested in recommendations.

excid3:
Thanks! What an easy mistake to make there.

test/delivery_methods/slack_test.rb
excid3: I think the best solution is if make the `logger` configurable so you can define your own and then we can define one just for testing like you did here.

mike-burns:
And you cut a release! That's amazingly fast. Thank you!

excid3:
You're welcome and thank you for the fix!

PR #300, opened 2023-07-27
Updating the email delivery method

tomrossi7:
This introduces a new option to the email delivery method to pass in arguments to mailers.

Sometimes a mailer will accept arguments:

```Ruby
class UserMailer < ApplicationMailer
  def comment_notification(comment)
    ...
  end
end
```

This can now be accounted for with an `args` option:

```Ruby
class EmailDeliveryWithArguments < Noticed::Base
  deliver_by :email, mailer: "UserMailer", method: :comment_notification, args: :email_args

  def email_args
    comment
  end
end
```

This will also work with a mailer that accepts named arguments with the `named_args` option:

```Ruby
class UserMailer < ApplicationMailer
  def comment_notification(comment:)
    ...
  end
end
```

```Ruby
class EmailDeliveryWithNamedArguments < Noticed::Base
  deliver_by :email, mailer: "UserMailer", method: :comment_notification_with, named_args: :email_named_args

  def email_named_args
    {comment: comment}
  end
end
```

### DIFF
--- a/lib/noticed/delivery_methods/email.rb
+++ b/lib/noticed/delivery_methods/email.rb
@@ -52,14 +52,22 @@ module Noticed
       end
 
       def args
-        return unless (option = options[:arguments])
+        return unless (option = options[:args])
+
+        notification.send(option)
+      end
+
+      def named_args
+        return unless (option = options[:named_args])
 
         notification.send(option)
       end
 
       def composed_mailer
-        if options[:arguments]
+        if options[:args]
           mailer.with(params).public_send(mailer_method.to_sym, args)
+        elsif options[:named_args]
+          mailer.with(params).public_send(mailer_method.to_sym, **named_args)
         else
           mailer.with(params).public_send(mailer_method.to_sym)
         end
--- a/lib/noticed/delivery_methods/email.rb
+++ b/lib/noticed/delivery_methods/email.rb
@@ -5,9 +5,9 @@ module Noticed
 
       def deliver
         if options[:enqueue]
-          mailer.with(format).send(mailer_method.to_sym).deliver_later
+          composed_mailer.deliver_later
         else
-          mailer.with(format).send(mailer_method.to_sym).deliver_now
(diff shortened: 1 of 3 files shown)


lib/noticed/delivery_methods/email.rb
```diff
@@ -50,6 +50,20 @@ def format
         end
         params.merge(recipient: recipient, record: record)
       end
+
+      def args
```
tomrossi7, line 54: I'm not sure if we should standardize on `args` or `arguments`? I usually don't like to abbreviate, but Its kind of like`params` and `parameters`?

test/dummy/app/mailers/user_mailer.rb
```diff
@@ -2,4 +2,8 @@ class UserMailer < ApplicationMailer
   def comment_notification
     mail(body: "")
   end
+
+  def comment_notification_for(user)
```
tomrossi7, line 6: I added a new mailer action that accepts an argument.

tomrossi7:
@excid3 what do you think about having options that just checking if a method is defined rather than passing it as an option? For example, rather than passing the option `arguments: :email_arguments` to the `delivery_by` method, what if we just check to see if `method_defined?(:email_arguments)`. This would codify the opinion that if you want to pass in arguments, you must define a method called `email_arguments`.

```
class EmailDeliveryWithArguments < Noticed::Base
  deliver_by :email, mailer: "UserMailer", method: :comment_notification_for, arguments: :email_arguments

  def email_arguments
    params[:recipient]
  end
end
```
would become:
```
class EmailDeliveryWithArguments < Noticed::Base
  deliver_by :email, mailer: "UserMailer", method: :comment_notification_for

  def email_arguments
    params[:recipient]
  end
end
```

You lose the flexibility to call the method whatever you want but gain convention over configuration goodness. Just a thought!

excid3:
This is rolled into v2. 👍

PR #304, opened 2023-08-10
More realistic APN sandbox support 

rromanchuk:
A user has_many tokens that can be generated from both development (sandboxed devices), or production (not sandboxed devices) and is unrelated to the rails environment or endpoint being used. It's per-token introspection, which makes it's it a little annoying juggling connection pools, but it's a life saver when correctly supported.

```ruby
  def ios_device_tokens(recipient)
    recipient.notification_tokens.where(environment: :production, platform: :iOS).pluck(:token)
  end

  def ios_development_device_tokens(recipient)
    recipient.notification_tokens.where(environment: :development, platform: :iOS).pluck(:token)
  end
```

### DIFF
--- a/lib/noticed/delivery_methods/ios.rb
+++ b/lib/noticed/delivery_methods/ios.rb
@@ -4,6 +4,7 @@ module Noticed
   module DeliveryMethods
     class Ios < Base
       cattr_accessor :connection_pool
+      cattr_accessor :development_connection_pool
 
       def deliver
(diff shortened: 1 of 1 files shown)


excid3:
I'd recommend using two separate iOS delivery methods for this case instead.

```ruby
deliver_by :ios do |config|
 config.device_tokens = ->(recipient) { retrieve production tokens }
end

deliver_by :ios_development do |config|
  config.development = true
  config.device_tokens = ->(recipient) { retrieve development tokens }
end
```

PR #305, opened 2023-08-11
Record failing request arguments

bvogel:
Currently when a POST to a service fails there is no information what parameter caused the failure, especially the device tokens of FCM delivery methods are missing. Adding the args to the exception allows to use `rescue_from Noticed::ResponseUnsuccessful` and have more meaningful data to react to the failure. 

Specific use case: I use FCM as delivery method and sometimes during testing device tokens from a different environment will end up in the database, however the response isn't a 404 and only tells me that the request failed due to a `SENDER_ID_MISMATCH`. With this change I'd be able to rescue from this exception and remove the offending device_token from the database.

Additionally I took the liberty to remove 2 `puts` from the production code.

### DIFF
--- a/lib/noticed.rb
+++ b/lib/noticed.rb
@@ -33,9 +33,11 @@ module Noticed
 
   class ResponseUnsuccessful < StandardError
     attr_reader :response
+    attr_reader :args
 
-    def initialize(response)
+    def initialize(response, args = {})
       @response = response
+      @args = args
     end
   end
 end
--- a/lib/noticed/delivery_methods/base.rb
+++ b/lib/noticed/delivery_methods/base.rb
@@ -85,7 +85,7 @@ module Noticed
         if !options[:ignore_failure] && !response.status.success?
           puts response.status
           puts response.body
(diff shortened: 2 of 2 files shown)


bvogel:
I'm really struggling without this, I keep getting 
```
exception: {
  message: "Noticed::ResponseUnsuccessful"
  name: "Noticed::ResponseUnsuccessful"
  stacktrace: [93]
}
```
in our k8s logs without any insights why the request is failing

stephaneliu:
Thanks for putting this together. Adding more insights to exception would greatly help our app as well.

Dev-Alchemist:
I fully support this. It will undoubtedly be beneficial.

bvogel:
Hey @excid3! First of all, thank you for this great gem, it really helps us a lot. I tried giving back, however it seems this PR went un"noticed". Mind sharing your thoughts? Or is there another way to get deeper insight into fails messages?

bvogel:
Thank you @excid3!

PR #311, opened 2023-10-12
Add support for HTML translations. Fixes #310

### LINKED ISSUE #310: Support html_safe translation
https://github.com/excid3/noticed/commit/dcaf132da5736e3356e256c240de8e804e952a5d added a `translate` helper, but it doesn't support keys (such as `message_html`) being automatically marked as HTML-safe. I could be a nice addition to simplify adding links to the notification message, for instance.

Relevant: https://github.com/rails/rails/blob/main/activesupport/lib/active_support/html_safe_translation.rb

excid3:
This allows you to use _html translations that will be automatically marked as HTML safe.

Fixes #310


excid3:
@Spone I think this should do the trick?

Spone (approved):
Nice!

Spone:
Maybe you need to require `active_support/html_safe_translation`?

excid3:
Hmm, looks like this module was added in Rails 7. 🙃

lib/noticed/translation.rb
```diff
@@ -1,7 +1,11 @@
+require "active_support/html_safe_translation"
```
Spone, line 1:
```suggestion
require "active_support"
require "active_support/html_safe_translation"
```

Maybe?

Spone:
> Alternatively, we could rescue from the require and only call super if it exists... which means this feature would only work for Rails 7+, but still can run on Rails 5.2 to 6.1.

Sounds good!

excid3:
Ayyy, there we go. Finally.

excid3:
Well shoot. Rafael said it's a private module, so it's not intended to be used like this. I guess we should reimplement ourselves in the gem.

PR #348, opened 2023-11-06
Discourage developers to check in their p8 file

joemasilotti:
When sending iOS notifications developers must provide their p8 certificate file. The default codepath looks in the Rails app/ directory which requires checking in the file to source control.

This PR moves the default location to the credentials file.

Note that it _is_ a breaking change for existing apps. I renamed the option to make it more clear what is expected.

We could keep the old method around but raise a deprecation warning when used. Or force a breaking change on folks to ensure the file isn't checked in.

### DIFF
--- a/docs/delivery_methods/ios.md
+++ b/docs/delivery_methods/ios.md
@@ -36,10 +36,6 @@ class CommentNotification
     apn.alert = "Hello world"
     apn.custom_payload = { url: root_url }
   end
-
-  def development?
-    Rails.env.local?
-  end
 end
 ```
 
--- a/docs/delivery_methods/ios.md
+++ b/docs/delivery_methods/ios.md
@@ -28,7 +28,7 @@ With custom configuration:
 
 ```ruby
 class CommentNotification
-  deliver_by :ios, format: :ios_format, development: :development?, pool_size: 5
+  deliver_by :ios, format: :ios_format, development: Rails.env.local?, pool_size: 5
 
   # Customize notification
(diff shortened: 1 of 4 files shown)


docs/delivery_methods/ios.md
```diff
@@ -28,7 +28,7 @@ With custom configuration:
 
 ```ruby
 class CommentNotification
-  deliver_by :ios, format: :ios_format, cert_path: :ios_cert_path, key_id: :ios_key_id, team_id: :ios_team_id, pool_size: 5
+  deliver_by :ios, format: :ios_format, development: :development?, pool_size: 5
```
joemasilotti, line 31: I use this `development?` on almost every project I work on so I think it is worth calling out here in the docs.

The two customizations were removed because they don't provide more value, in my opinion, over `#ios_format` customized above.

excid3: You can probably just do this inline since the Rails env will not change.

```suggestion
  deliver_by :ios, format: :ios_format, development: Rails.env.local?, pool_size: 5
```

excid3:
I think this looks good. We will need a CHANGELOG entry for this and it will be a breaking change, so probably requires a major release.

joemasilotti:
Would you prefer to do a minor release with a deprecation warning? We are technically changing the property name - using the old (existing) one could log a warning.

joemasilotti:
@excid3, I added a CHANGELOG entry and merged in `main`. Let me know if there's anything else!

PR #362, opened 2024-01-15
Add notification_methods for adding methods to the Notification model

excid3:
This introduces a `notification_methods` block for Notifiers to add methods to the Noticed::Notification records. This works by adding STI to Noticed::Notifications, dynamically creating a Notification class for each Notifier and applying methods to it.

Without this feature, it is tricky to load notifications from the database and call methods on the Notifier that require the `notification` or `recipient`:

```erb
<div>
  <% @user.notifications.includes(:event).each do |notification| %>
    <%= link_to notification.event.message_for(notification)  %>
  <% end %>
</div>
```

```ruby
class NewCommentNotifier < Noticed::Event
  # ...

  def message_for(notification)
    "#{notification.recipient.name}, you have a new comment to review"
  end
end
```

While this is fine, it feels incomplete.

By introducing `notification_methods`, we can refactor the above to this:

```erb
<div>
  <% @user.notifications.includes(:event).each do |notification| %>
    <%= link_to notification.message  %>
  <% end %>
</div>
```

```ruby
class NewCommentNotifier < Noticed::Event
  # ...

  notification_methods do
    def message
      "#{recipient.name}, you have a new comment to review"
    end
end
```

This is considerably cleaner and provides a handy way of extending the Notification class while also making the methods unique to each Notifier. If we added the `message` method to `Noticed::Notification` directly, it would share that method across all Notifiers.

See https://github.com/excid3/noticed/discussions/360 for the discussion leading to this.

### DIFF
--- a/app/models/concerns/noticed/deliverable.rb
+++ b/app/models/concerns/noticed/deliverable.rb
@@ -130,6 +130,7 @@ module Noticed
 
     def recipient_attributes_for(recipient)
       {
+        type: "#{self.class.name}::Notification",
         recipient_type: recipient.class.name,
         recipient_id: recipient.id
       }
--- a/app/models/concerns/noticed/notification_methods.rb
+++ b/app/models/concerns/noticed/notification_methods.rb
@@ -1,0 +1,17 @@
+module Noticed
+  module NotificationMethods
+    extend ActiveSupport::Concern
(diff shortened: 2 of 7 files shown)


jon-sully:
🙌

PR #377, opened 2024-01-23
iOS Delivery Method Updates and a couple of docs enhancements

phil-6:
### Functionality

- Changes `format` to `json` in `lib/noticed/delivery_methods/ios.rb`
- Allows `json` to respond to a symbol as a method in `ios.rb`
- Allows connection_pool_options to respond to symbol representing a method

### Documentation
- Add section about delivery method configuration to README.md
- Add tip about Extracting Delivery Method Configurations to README.md
- Update UPGRADE.md with changes to configuration methods

### Changes `format` to `json` in `lib/noticed/delivery_methods/ios.rb`

The UPGRADE.md lists one of the changes being format option for both iOS and FCM being renamed to JSON. This change had only been implemented for FCM. I'm not certain the approach I've taken here is the right one, it might be better to keep the code as-is and revert the change in the docs. It depends what the original intent was.

### Allows `json` to respond to a symbol as a method in `ios.rb`

This change is `lib/noticed/delivery_methods/ios.rb#L36` and adds:
``` ruby
method = event.send(method, apn) if method.is_a?(Symbol) && event.respond_to?(method)
```

Using the same method as `evaluate_options()` in the DeliveryMethod class. It evaluates the `method` if it's a symbol and event responds to that method.

Note: I think a similar change might be needed in fcm.rb, but the execution logic in `send_notification(device_token)` is different and I can't get it working. 

### Allows connection_pool_options to respond to symbol representing a method

Similar to the change above, but using the `evaluate_options()` method to allow extraction of values to methods

## Testing
All tests pass.

## Checklist

- [x] All existing tests pass

---

Note: This is my first open source code PR. Happy to take on any and all feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ To start, generate a Notifier:
 rails generate noticed:notifier NewCommentNotifier
 ```
 
-#### Notifier Objects
+### Notifier Objects
 
 Notifiers are essentially the controllers of the Noticed ecosystem and represent an Event. As such, we recommend naming them with the event they model in mind — be it a `NewSaleNotifier,` `ChargeFailureNotifier`, etc.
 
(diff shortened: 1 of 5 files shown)


docs/delivery_methods/ios.md
```diff
@@ -22,7 +22,7 @@ https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_
 class CommentNotification
   deliver_by :ios do |config|
     config.device_tokens = ->(recipient) { recipient.notification_tokens.where(platform: :iOS).pluck(:token) }
-    config.format = ->(apn) {
```
excid3, line 25: Since we're using the Apnotic gem and not an HTTP request, I figured this shouldn't be renamed to json. I copied the line in the UPGRADE guide on accident.

phil-6: Yep, that makes sense. I'll revert that change and update the docs accordingly

README.md
```diff
@@ -117,7 +117,75 @@ end
...
+```ruby
+class CommentNotifier < Noticed::Event
+  deliver_by :ios do |config|
+    config.format = :ios_format
+    config.apns_key = :ios_cert
+    config.key_id = :ios_key_id
+    config.team_id = :ios_team_id
+    config.bundle_identifier = :ios_bundle_id
+    config.device_tokens = :ios_device_tokens
+    config.if = :send_ios_notification?
```
excid3, line 138: Should we show a mixture here of methods, lambdas, and values?

```suggestion
    config.bundle_identifier = Rails.application.credentials.dig(:ios, :bundle_identifier)
    config.device_tokens = :ios_device_tokens
    config.if = ->(notification) { recipient.send_ios_notification? }
```

phil-6: More examples more better!

PR #382, opened 2024-01-23
Add ephemeral notifiers

excid3:
This adds an Ephemeral class for sending notifications. This was previously a feature of Noticed v1 but was removed to make solve for bulk notifications. The way things are architected, we can simulate the Notifier & Notification classes for databaseless / ephemeral notifications that should not be saved in the database.

Ephemeral notifiers basically just pass the params and recipients to the delivery method jobs directly. The Notifier and Notification use ActiveModel to set up the same basic attributes as the database models in order to achieve compatibility. It seems to work pretty well.

This is an experimental idea to see if it would be a reasonable solution. There are limitations to this in that config options must be serializable by ActiveJob (meaning Procs and Lambdas cannot be used for delivery method configs, etc).

See #375


excid3:
@viktor-shmigol have you had a chance to test this?

cjilbert504 (approved):

viktor-shmigol:
@excid3 I'm sorry for the long response. I was a little bit busy.
I've tested it and I found that I'm not able to call  any notifier method from the Noticed::DeliveryMethod instance.
For example:
```
class NewMessageNotifier < Noticed::Ephemeral
  deliver_by :internal_notification, class: 'DeliveryMethods::InternalNotification'

  required_params :message

  def channel
    "user_#{recipient.id}"
  end

  def payload
    { body: }
  end

  def body
    I18n.t('notifications.new_message', author: params[:message].author_name)
  end
end


class DeliveryMethods::InternalNotification < Noticed::DeliveryMethod
  delegate :event, to: :notification

  def deliver
    ActionCable.server.broadcast(event.channel, event.payload)
  end
end

NewMessageNotifier.new(params: {message: Message.last}).deliver(User.last)
```

I'm getting this error:

Error performing DeliveryMethods::InternalNotification (Job ID: 0833436a-bbcd-4ee3-811f-ce31362c6d98) from Async(default) in 9.52ms:
NoMethodError (undefined method `channel' for #<Noticed::Ephemeral:0x0000ffff8149cb50 @attributes=#<ActiveModel::AttributeSet:0x0000ffff8149c8a8 @attributes={"params"...

If I wrap those methods in notification_methods:

```
class NewMessageNotifier < Noticed::Ephemeral
  deliver_by :internal_notification, class: 'DeliveryMethods::InternalNotification'

  required_params :message

  notification_methods do
    def channel
      "user_#{recipient.id}"
    end

    def payload
      { body: }
    end

    def body
      I18n.t('notifications.new_message', author: params[:message].author_name)
    end
  end
end
```
I'm getting this error:

NoMethodError: undefined method `notification_methods' for NewMessageNotifier:Class

Also, it would be good to have a way to deliver notifications in foreground mode using perform_now.
E.g:
NewMessageNotifier.with({message: Message.last}).deliver_now(User.last)

excid3:
@viktor-shmigol Give this a try again. It was retrieving `Noticed::Ephemeral::Notification` constant instead of `NewMessageNotifier::Notification` which is what caused the method missing error.

bvogel:
I know this is an old PR, but I seem to be missing something: 
we have user options that allow our users to determine which kind of notifications they will receive. With v1 there was a setting for web/database and a setting for push/fcm. So the combinations are both (web/push), only web (web/-), only push (-/push) and none at all (-/-) With the v2 options I can cover easily "both", "only web" and "none" - however for "only push" I'd need an ephemeral notifier - does that mean I need to double the code for all notifiers (currently 14 different ones) and select either the regular or the ephemeral based on the users setting? Is that how this is supposed to work? TBH the previous `if:` for the database method was much more serviceable.

excid3:
@bvogel Please open up a discussion with an example to showcase what you're trying to do.

bvogel:
thanks: https://github.com/excid3/noticed/discussions/472

Issue #383, opened 2024-01-24
[BUG] Inaccurate recipient_type class name for STI model, when delivering the notification


alikrugl:
## Bug Report

**Describe the Bug:**

When delivering notifications to the recipient of the STI class, the `recipient_type` will be set to the class name of the STI model, _instead_ of the base model name. Due to this issue, the association `:notifications` may not function correctly for STI records. Currently, an empty collection is being returned, as the `recipient_type` is checked against the **base model name** instead of the **STI class**.

 It's important to note that according to the Ruby on Rails API documentation the polymorphic `type` column should store the **base model class** (refer to: https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#:~:text=Using%20polymorphic%20associations,the%20posts%20table.)

Related code is located on the line `app/models/concerns/noticed/deliverable.rb:94`
https://github.com/excid3/noticed/blob/main/app/models/concerns/noticed/deliverable.rb#L94
```ruby
recipient_type: recipient.class.name,
``` 

**To Reproduce:**

1. Deliver the notification to the STI model record

    ```ruby
    CommentNotifier.with(record: comment).deliver(UserSti.last)
    ``` 
    `Noticed::Notification` record will be filled with `recipient_type: 'UserSti'`
2. Fetch the notifications from the STI record.
    ```ruby
    UserSti.last.notifications # => []
    ``` 
    It will return an empty array because within the query it searches for the `User` `recipient_type` `"noticed_notifications"."recipient_type"="User"`

**Expected Behavior:**
`recipient_type` of the `Noticed::Notification` record should match the base class of the `recipient` record

**Actual Behavior:**
`recipient_type` of the `Noticed::Notification` record matches the STI class name of the `recipient` record

**Screenshots (if applicable):**
![image](https://github.com/excid3/noticed/assets/71523527/899b9f12-fdc1-4776-bd47-86f90085d416)

**Environment:**
- Noticed gem version: 2.0.3
- Ruby version: 3.2.2
- Rails version: 7.0.8
- Operating System:  Ubuntu 20.04

**Possible Fix:**
Use `base_class` when specifying `recipient_type`
```ruby
recipient_type: recipient.class.base_class.name,
``` 

**Steps to Reproduce with Fix (if available):**

**Related Issues:**

**Labels to Apply:**
bugfix

**Checklist:**

- [x] I have searched for similar issues and couldn't find any
- [x] I have checked the documentation for relevant information
- [x] I have included all the required information

PR #384, opened 2024-01-24
Use base_class when specifying recipient_type

alikrugl:
## Pull Request

**Summary:**
Updates `app/models/concerns/noticed/deliverable.rb` to ensure it stores the base model class for STI models in the `recipient_type` column of the polymorphic association.

**Related Issue:**
Fixed https://github.com/excid3/noticed/issues/383

**Testing:**
All Tests are passing
I have installed the feature in my app and tested it.

**Checklist:**

- [x] All existing tests pass

***

This is my first time contributing to open source, so any feedback is appreciated!

### DIFF
--- a/app/models/concerns/noticed/deliverable.rb
+++ b/app/models/concerns/noticed/deliverable.rb
@@ -91,7 +91,7 @@ module Noticed
     def recipient_attributes_for(recipient)
       {
         type: "#{self.class.name}::Notification",
-        recipient_type: recipient.class.name,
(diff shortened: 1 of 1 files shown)


excid3:
I think this will be a good default. I would assume most STI models as recipients will have notifications associated on the base class. However, it's possible that they might not and this would not work in that situation. Those cases can be solved by overriding the `recipient_attributes_for` method to ignore the base class.

We need to add tests and documentation for this.

PR #390, opened 2024-01-26
Returning the result of the symbol evaluation for ios format

avanrielly:
## Pull Request

**Summary:**
Making sure to return after evaluating an iOS symbolized format method.

**Related Issue:**

**Description:**
Right now, the `Noticed::DeliveryMethods::Ios#format_notification` method is setup to handle a proc, symbol or message.  When the symbol version of the format method is used, it evaluates the symbol using send and then calls `notification.instance_exec` with the result from the `event.send` call.  This causes an error in execution.

**Testing:**
This was tested to work on my local machine against our codebase.

**Checklist:**

- [x] All existing tests pass

**Additional Notes:**

### DIFF
--- a/lib/noticed/delivery_methods/ios.rb
+++ b/lib/noticed/delivery_methods/ios.rb
@@ -35,7 +35,7 @@ module Noticed
         method = config[:format]
         # Call method on Notifier if defined
         if method&.is_a?(Symbol) && event.respond_to?(method)
-          return event.send(method, apn)
+          event.send(method, apn)
         # If Proc, evaluate it on the Notification
         elsif method.present?
           notification.instance_exec(apn, &method)
--- a/lib/noticed/delivery_methods/ios.rb
+++ b/lib/noticed/delivery_methods/ios.rb
@@ -33,7 +33,7 @@ module Noticed
         apn.topic = evaluate_option(:bundle_identifier)
 
         if (method = config[:format])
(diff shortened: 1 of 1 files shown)


excid3:
Ah yes, I missed this when reviewing!

cc @phil-6

lib/noticed/delivery_methods/ios.rb
```diff
@@ -32,8 +32,12 @@ def deliver
...
         apn.topic = evaluate_option(:bundle_identifier)
 
-        if (method = config[:format])
-          method = event.send(method, apn) if method.is_a?(Symbol) && event.respond_to?(method)
+        method = config[:format]
+        # Call method on Notifier if defined
+        if method&.is_a?(Symbol) && event.respond_to?(method)
+          event.send(method, apn)
+        # If Proc, evaluate it on the Notification
+        elsif method.present?
```
avanrielly, line 40: What are your thoughts on having this mirror the `evaluate_options` call? , where it checks `method.respond_to(:call)` and then falls back to the string value of method if it exists?  This way the iOS format would match the way `evaluate_options` works for custom delivery methods with the final fallback to `message` if it exists.

excid3: Yes, that's what I intended actually. 
```suggestion
        elsif method&.respond_to?(:call)
```

PR #398, opened 2024-02-08
Example in generator should have the correct name

emilkarl:
Generated a Notifier and the example comment in the top was wrong, missing `Notifier`.


excid3:
Thank you!

PR #399, opened 2024-02-09
Make the ApiClient use application/x-www-form-urlencoded as content type

supherman:
## Summary
One of the issues I encountered after upgrading to V2 is that all of the tests that were stubbing requests to twilio failed, turns out Webmock does not support matching the request body for multipart/form-data requests, the error thrown was this:
 
```
ArgumentError: WebMock does not support matching body for multipart/form-data requests yet :(
```

I believe `multipart/form-data` is mostly used for file uploads, so, perhaps using `application/x-www-form-urlencoded ` makes more sense to submit plain form data.

## Testing
All Tests are passing, added body matching to the Twilio request stub.

## Checklist
- [x] All existing tests pass

### DIFF
--- a/lib/noticed/api_client.rb
+++ b/lib/noticed/api_client.rb
@@ -28,7 +28,7 @@ module Noticed
       if (json = args.delete(:json))
         request.body = json.to_json
       elsif (form = args.delete(:form))
-        request.set_form(form, "multipart/form-data")
+        request.form_data = form
       end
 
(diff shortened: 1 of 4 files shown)


excid3:
I think this is what I intended originally but wasn't paying close attention to it.

Chances are you won't be uploading files in notifications and you can always implement your own POST request if you did want to, so this makes sense to me to update.

PR #402, opened 2024-02-14
Honor PK settings

rromanchuk:
## Pull Request

**Summary:**
This was talked about in the README, but I missed it from the upgrading doc. I should know better since I make this mistake/forget every time I run a gem migrations. 

I just copied ActiveStorage style https://github.com/rails/rails/pull/42378/files#diff-a99d3a53b13fcab0bb710064be91c98c5929b3ecc8a000d6b9af47ec74a4c6c4 

For those that utilize
```ruby
# application.rb
config.generators do |g|
  g.orm :active_record, primary_key_type: :uuid
end
```


excid3:
This would be handy to have built-in (and for any gem with migrations).

PR #405, opened 2024-02-14
Minor sig fix in docs

rromanchuk:
Update yield example, there might be a few more spots


README.md
```diff
@@ -348,7 +348,7 @@ module IosNotifier
       config.key_id = Rails.application.credentials.dig(:ios, :key_id)
       config.team_id = Rails.application.credentials.dig(:ios, :team_id)
       config.apns_key = Rails.application.credentials.dig(:ios, :apns_key)
-      config.if = ->(recipient) { recipient.ios_notifications? }
+      config.if = ->(notification) { notification.recipient.ios_notifications? }
```
excid3, line 351: Is this right? Shouldn't it be
```suggestion
      config.if = -> { recipient.ios_notifications? }
```

PR #419, opened 2024-03-02
Add the `if` and `unless` config check before enqueueing a job so that unnecessary jobs can be avoided

justwiebe:
## Pull Request

**Summary:**

Updated the EventJob to check the DeliverBy `if`/`unless` conditionals so that no unnecessary jobs are queued

**Related Issue:**

**Description:**

My project needs to send out notifications to 7000 users at a time. We are adding mobile app notifications, but very few users will have the mobile app, and even less should need iOS *and* Android notifications. Currently the `if`/`unless` conditionals are run after the job for that delivery method is queued up, which lets the job no-op, but with 14,000 jobs created, it can take a long time to run through them, even if they aren't doing anything.

To save time and allow the real notifications to be processed sooner, we shouldn't create jobs if we don't have to.

**Testing:**

I created unit tests on the EventJob to make sure that no extra jobs are queued up.

**Checklist:**

- [x] All existing tests pass

**Additional Notes:**

### DIFF
--- a/app/jobs/noticed/event_job.rb
+++ b/app/jobs/noticed/event_job.rb
@@ -4,14 +4,14 @@ module Noticed
 
     def perform(event)
       # Enqueue bulk deliveries
-      event.bulk_delivery_methods.each do |_, deliver_by|
+      event.bulk_delivery_methods.each_value do |deliver_by|
         deliver_by.perform_later(event)
       end
 
       # Enqueue individual deliveries
       event.notifications.each do |notification|
-        event.delivery_methods.each do |_, deliver_by|
-          deliver_by.perform_later(notification)
+        event.delivery_methods.each_value do |deliver_by|
+          deliver_by.perform_later(notification) if deliver_by.perform?(notification)
         end
       end
     end
--- a/app/models/noticed/deliverable/deliver_by.rb
+++ b/app/models/noticed/deliverable/deliver_by.rb
@@ -19,33 +19,41 @@ module Noticed
       end
 
       def perform_later(event_or_notification, options = {})
-        options[:wait] ||= evaluate_option(:wait, event_or_notification) if config.has_key?(:wait)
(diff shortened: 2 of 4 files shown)


excid3:
The jobs must run the conditions at execution time.

The reason being that a notification may be delivered several different ways:

Send websocket notification immediately
If unread after 2 minutes, send push notification
If unread after 10 minutes, send email notification

These need to evaluate during the execution of the job, not before.

justwiebe:
Okay, I'll update it. What do you think of `queue_delivery_job?` as a method name @excid3?

excid3:
> What do you think of `queue_delivery_job?` as a method name @excid3?

I don't think it clearly describes it. Since it will always delivery by default, what if it was something like `config.skip_delivery_if`

excid3:
I am still finding it hard to differentiate between `if` after sleeping on this. I was just reading Andy Croll's blog post and debating on naming still. https://andycroll.com/ruby/a-job-should-know-whether-to-run-itself/

`skip_enqueue_if` might be a better name? 

```ruby
class MessageNotifier
  deliver_by :action_cable

  deliver_by :email do |config|
    config.unless = ->{ read_at? }
    config.skip_enqueue_if = ->{ recipient.

**Comment by justwiebe** [comment]
What would your thoughts be on using `skip_enqueue_if` and `notify_if` instead of `if` and `notify_unless` instead of `unless`? Obviously those last two would need to be additional, with `if`/`unless` deprecated, unless you want to go to version 3 already

**Comment by excid3** [comment]
Noodling on it, I'm thinking `before_enqueue` may be good at providing additional functionality as a callback along with the ability to abort enqueues. I took a stab at refactoring that last night and it seems good.

```ruby
config.before_enqueue = ->{ throw(:abort) unless recipient.email_notifications? }
```

justwiebe:
Great, thanks for doing that!

excid3:
Thanks for this PR @justwiebe! 🙏

PR #429, opened 2024-03-20
Misc fixes to allow dynamic mailer name and dynamic wait while sendin…

ausangshukla:
…g emails

## Pull Request

**Summary:**
These changes enable for mailer and wait to take method names as symbols

**Related Issue:**

**Description:**
Its a pretty simple change
1. mailer = evaluate_option(:mailer), to enable the mailer to be dynamic and based on a method defined in the event.
2. option.is_a?(Symbol) && context.event.respond_to?(option), allows the wait time to be dynamic and based on a method call.

**Testing:**
Ive run my applications tests against this, but have not written any new tests specifically for these changes.

**Checklist:**

- [ ] All existing tests pass

**Additional Notes:**

### DIFF
--- a/app/models/noticed/deliverable/deliver_by.rb
+++ b/app/models/noticed/deliverable/deliver_by.rb
@@ -34,6 +34,8 @@ module Noticed
           context.instance_exec(&option)
         elsif option.is_a?(Symbol) && context.respond_to?(option)
           context.send(option)
+        elsif option.is_a?(Symbol) && context.event.respond_to?(option)
+          context.event.send(option)
         else
           option
         end
--- a/lib/noticed/delivery_methods/email.rb
+++ b/lib/noticed/delivery_methods/email.rb
@@ -4,7 +4,7 @@ module Noticed
       required_options :mailer, :method
 
       def deliver
-        mailer = fetch_constant(:mailer)
+        mailer = evaluate_option(:mailer)
         email = evaluate_option(:method)
         args = evaluate_option(:args) || []
         mail = mailer.with(params).send(email, *args)


excid3:
I actually just commited a fix here: https://github.com/excid3/noticed/commit/8da57f851914d30ddd6ee773e06274bf0ef4fdcc

Is there a case this doesn't handle?

ausangshukla:
This may work for the dynami mailer, but not for the wait time.

excid3:
Can you provide an example?

ausangshukla:
This is what I have in my actual prod code. The reason is to prevent aws from getting flooded, which results in throttling errors. So we need to delay the sending of emails. However when sending a bunch of emails, we cant use the same delay, we need to spread the delay over some period, by randomizing it, so that aws does not get flooded. Hope this makes sense

deliver_by :email do |config|
      config.mailer = :mailer_name
      config.method = :email_method
      config.params = :email_data
      config.wait = :email_delay
 end


def email_delay(_notification = nil)
    # Randomize the delay so we dont flood aws SES
    delay_seconds = entity.entity_setting.email_delay_seconds
    delay_seconds = delay_seconds.positive? ? rand(1..delay_seconds) : rand(1..300)
    Rails.env.development? ? 0 : delay_seconds
 end

ausangshukla:
I agree with your changes in https://github.com/excid3/noticed/commit/8da57f851914d30ddd6ee773e06274bf0ef4fdcc, can replace the one I suggested, for dynamic mailer

ausangshukla:
Also note that in version < 2 I could do the following
deliver_by :email, mailer: :mailer_name, method: :email_method, format: :email_data, delay: :email_delay

And the delay would be dynamic, this is the prob Im trying to solve above.

excid3:
Wait should already work, but it will be evaluated against methods in the `notification_methods` block since it would apply to the individual recipients.

ausangshukla:
Makes sense, thank you for clarifying. Appreciate it.

PR #439, opened 2024-04-18
Handle uncaught 'Unregistered' not being caught by 410 status code

rromanchuk:
## Pull Request

**Summary:**
Number one uncaught fatal right now is from sidekiq retrying to to deliver APNs that does not catch my token clean up proc, so it will continue to retry until retired into dead queue.  

**Description:**
Raise with http status code for additional debugging context. `response.status == "410"` is currently **not** catching it. Consider a `bad_token?` true if  `response.body["reason"]` equals 'Unregistered'. I have no idea at the moment, why the status code isn't catching it.  The error doesn't surprise me because of the velocity of different bundle ids/entitlements/environments i'm testing with. Just 4 users (most likely all me)  responsible >3k+ errors (retry policy), but is never able to clean up.  

I read this three times and have no idea what it means, but seems like it's probably responsible
> Table 3 lists the keys found in the JSON dictionary for unsuccessful requests. The JSON data might also be included in a GOAWAY frame when a connection is terminated.

https://developer.apple.com/documentation/usernotifications/handling-notification-responses-from-apns#Understand-error-codes

**Testing:**
None yet, will be pushing straight to production to snag the actual http status code 

Screenshot 2024-04-17 at 9 47 28 PM

Screenshot 2024-04-17 at 9 48 40 PM

**Checklist:**

- [ ] All existing tests pass

**Additional Notes:**

### DIFF
--- a/lib/noticed/delivery_methods/ios.rb
+++ b/lib/noticed/delivery_methods/ios.rb
@@ -50,8 +50,6 @@ module Noticed
         response.status == "410" || (response.status == "400" && response.body["reason"] == "BadDeviceToken") || response.body["reason"] == "Unregistered"
       end
 
-      def unregistered?(response)
-
       def development_pool
         self.class.development_connection_pool ||= new_connection_pool(development: true)
       end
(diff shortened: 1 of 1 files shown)


rromanchuk:
This is invalid, method symbol/proc was never passed via config,  so `config[:invalid_token]` was why it was branching. Closing for user error, must have missed it when migrating.

PR #444, opened 2024-05-19
Twilio delivery method allows Notifier to handle errors

bibstha:
## Pull Request

**Summary:**
This PR allows handling Twilio errors in the Notifier. This is very similar to how other DeliveryMethods do it.
eg in fcm.rb

https://github.com/excid3/noticed/blob/1514ae3f239b991a3b952e3805474c147e68ea4e/lib/noticed/delivery_methods/fcm.rb#L18-L23

**Description:**
Twilio has very verbose errors (including error codes) that allows API callers to do follow up actions.
eg: Error code [21211 ](https://www.twilio.com/docs/api/errors/21211) shows that the number is invalid. In this case, we can mark a phone number as invalid to prevent further calls to Twilio.

**Testing:**
Added a test for the added code.

Also added example in the Twilio documentation

Screenshot 2024-05-19 at 12 15 52 PM

**Checklist:**

- [x] All existing tests pass

**Additional Notes:**

### DIFF
--- a/docs/delivery_methods/twilio_messaging.md
+++ b/docs/delivery_methods/twilio_messaging.md
@@ -38,6 +38,23 @@ deliver_by :twilio_messaging do |config|
 end
 ```
 
+## Error Handling
+
+```ruby
+deliver_by :twilio_messaging do |config|
+  config.error_handler = lambda do |twilio_error_response|
+    error_hash = JSON.parse(twilio_error_response.body)
+    case error_hash["code"]
+    when 21211
+      # The 'To' number is not a valid phone number.
+      # Write your error handling code
+    else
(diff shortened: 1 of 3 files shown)


docs/delivery_methods/twilio_messaging.md
```diff
@@ -38,6 +38,23 @@ deliver_by :twilio_messaging do |config|
 end
 ```
 
+## Error Handling
+
```
excid3, line 42:
```suggestion

Twilio provides a full list of error codes that can be handled as needed. See https://www.twilio.com/docs/api/errors

```

excid3:
I think this makes sense. I don't imagine you'll need to handle anything other than 4xx errors.

PR #459, opened 2024-06-26
Add recipients feature for defining recipients inside the notifier

excid3:
Currently, Noticed requires you to precompute the recipients of a Notifier.

```ruby
recipients = comment.post.commenters.excluding(comment.user).uniq
CommentNotifier.with(comment: @comment).deliver(recipients)
```

Instead, we could move this into the Notifier and compute it internally

```ruby
class CommentNotifier < ApplicationNotifier
  recipients ->{ params[:comment].post.commenters.excluding(params[:comment].user).distinct }

  # or 
  recipients do 
    params[:comment].post.commenters.excluding(params[:comment].user).distinct 
  end

  # or
  recipients :fetch_recipients

  def fetch_recipients
     # ...
  end
end
```

This greatly simplifies the code required in controllers or other locations where Notifiers are triggered.
```ruby
CommentNotifier.with(comment: @comment).deliver
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,21 @@
 ```ruby
 class CommentNotifier < ApplicationNotifier
   # Notify all the commenters on this post except the new comment author
-  recipients ->{ params[:comment].post.commenters.uniq - params[:comment.user] }
+
+  # Can be given a lambda or Proc
+  recipients ->{ params[:comment].post.commenters.excluding(params[:comment].user).distinct }
+
+  # Can be given a block
+  recipients do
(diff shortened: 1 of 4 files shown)


misterhtmlcss:
Is this better than using this? Just wondering if there are advantages I’m not aware of.

```ruby
class CommentNotifier < ApplicationNotifier
  def initialize(comment)
    @comment = comment
  end

  def recipients
      @comment.post.commenters.excluding(@comment.user).uniq
  end

  def deliver
    # implementation here
  end
end

CommentNotifier.new(@comment).deliver
```

excid3:
@misterhtmlcss I'm not following?

PR #470, opened 2024-07-15
Fix undefined ApplicationDeliveryMethod

adrienpoly:
## Pull Request

**Summary:**

While trying to implement a custom delivery method with Noticed V2 we had to replace `ApplicationDeliveryMethod` by `Noticed::DeliveryMethod`

I could find any reference in the code to ApplicationDeliveryMethod so I think this should be the correct way of doing it 

**Checklist:**

- [ ] All existing tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -525,10 +525,10 @@ If you want to build your own delivery method to deliver notifications to a spec
 
 `rails generate noticed:delivery_method Discord`
 
-This will generate a new `DeliveryMethods::Discord` class inside the `app/notifiers/delivery_methods` folder, which can be used to deliver notifications to Discord.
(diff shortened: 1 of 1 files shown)


excid3:
The generator creates `ApplicationDeliveryMethod` https://github.com/excid3/noticed/blob/main/lib/generators/noticed/delivery_method_generator.rb#L15

adrienpoly:
ah ok thanks I didn't use the generator to create this custom delivery method. 

Maybe then we could update the texte like that 

```diff
- This will generate a new DeliveryMethods::Discord class inside the app/notifiers/delivery_methods folder, which can be used to deliver notifications to Discord.
+ This will generate a new ApplicationDeliveryMethod and DeliveryMethods::Discord class inside the app/notifiers/delivery_methods folder, which can be used to deliver notifications to Discord.
```

adrienpoly:
done

PR #483, opened 2024-08-28
Allow changing the base class for `EventJob`

ghiculescu:
This is necessary if you want to change the queue for `EventJob`. Otherwise https://github.com/excid3/noticed/discussions/192#discussioncomment-1959422 doesn't work.

### DIFF
--- a/app/jobs/noticed/event_job.rb
+++ b/app/jobs/noticed/event_job.rb
@@ -1,7 +1,5 @@
 module Noticed
-  class EventJob < ApplicationJob
-    queue_as :default
-
+  class EventJob < Noticed.parent_class.constantize
     def perform(event)
       # Enqueue bulk deliveries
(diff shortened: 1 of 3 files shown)


app/jobs/noticed/event_job.rb
```diff
@@ -1,7 +1,5 @@
 module Noticed
-  class EventJob < ApplicationJob
-    queue_as :default
```
ghiculescu, line 3: This is [already](https://github.com/rails/rails/blob/a11f0a63673d274c59c69c2688c63ba303b86193/activejob/lib/active_job/queue_name.rb#L9) the default queue name, but by specifying it here, it means users can't change it at the Rails config level.

excid3:
Accidentally lost this one when I rewrote the new version. Appreciate the help!

PR #484, opened 2024-09-05
Include `Noticed::Translation` in `Noticed::Ephemeral::Notification`

ghiculescu:
This makes it more compatible with https://github.com/excid3/noticed/blob/b2f4e3cf9f377b93628497ec0c0bfd77fcef63af/app/models/noticed/notification.rb#L5


excid3:
Good catch @ghiculescu 👍

PR #485, opened 2024-09-06
Include  routes and `Noticed::Translation` in `Noticed::Ephemeral`

ghiculescu:
Ooops, I should have included this in https://github.com/excid3/noticed/pull/484, sorry.

Matches https://github.com/excid3/noticed/blob/505e6c9fc0cc58eb27714ddd449c819490f57368/app/models/noticed/event.rb#L5


excid3:
Ah yep, missed that too.

I also just realized, I don't think we include the url_helpers or translation in Event::Notification which we probably should do too.

PR #486, opened 2024-09-10
Fix crash when trying to bulk notify from an ephemeral notifier

ghiculescu:


ghiculescu:
Does this look okay to you @excid3

excid3:
Bulk delivery methods still seem incorrect when given recipients, so they'll need some more work to clean up.

PR #492, opened 2024-09-23
Allow use of evaluated recipients in ephemeral notifiers

ghiculescu:


excid3:
Thanks! I noticed this on Friday too but didn't get to finish it because it turned into a bit of a refactor.

PR #494, opened 2024-10-11
Update Documentation: Add details about how to vary queue priority

phil-6:
**Summary:**

Updates the documentation to add details about how to vary the queue based on the event. 

It took me a while to figure out that the `queue` config option is evaluated in the context of the notification in `app/models/noticed/deliverable/deliver_by.rb` rather than in the context of the event.


excid3:
Thanks @phil-6. I refactored this a bit to be 2 separate blocks of options, 1 for if/else which runs during the job, and the others which configure the job.

For anyone coming across this, see: https://github.com/excid3/noticed/blob/v2.4.3/app/models/noticed/deliverable/deliver_by.rb#L55-L61

PR #512, opened 2024-12-15
Documentation Update

d3rt:
## Pull Request

**Summary:**
Not a mistake, rather an optimization of the documentation.

**Related Issue:**
None

**Description:**
I just thought for 5 minutes about whether it might be clearer with the code block underneath, where you said "Noticed::NotificationExtensions".

**Testing:**
None.

None.

**Checklist:**
- [x] All existing tests pass

**Additional Notes:**
Maybe it will help a beginner like me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,7 +474,7 @@ Rails.application.config.to_prepare do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


excid3:
Good catch. I think I'm going to do the opposite and update the module name instead so it's clearer that you can use any name and it doesn't have to be inside the `Noticed` namespace.

PR #514, opened 2024-12-16
Inheritable notification methods

excid3:
When inheriting Notifiers, the `Notification` class will also inherit from it's parent's `Notification` class.

Discussion #509

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### Unreleased
 
+* Inherit Notification from parent class Notification.
 * Support callbacks in bulk delivery methods.
 
 ### 2.4.3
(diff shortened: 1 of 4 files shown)


excid3:
@wdiechmann this should add the inheritance we were talking about in #509 👍

wdiechmann:
Christmas did indeed come early this year 🎄 

huge thx @excid3 😍 

(and with such elegance if I may say so)

wdiechmann:
Updated and deploy'ed to production with warm thanks and best wishes for your seasons holidays and a productive/joyful New Year 😍

excid3:
@wdiechmann you too! 🎄 🎅 🤶 🎁

PR #515, opened 2024-12-28
Slack notifiers: add option to raise exception on failed 2xx responses

sherifnada:
## Pull Request

**Summary:**
Adds an option `raise_on_failure` to Slack notifiers which raises an exception if Slack returns an HTTP request with a 2xx status code but the `ok: false` body parameter. 

The param defaults to `false` to maintain backwards compatibility. 

**Related Issue:**
N/A

**Description:**
I faced an issue where the Slack bot I was using to deliver a message wasn't invited to the channel where it was trying to post. The Slack API was returning a 200 response but with the `ok: false` body parameter. This caused a silent failure. So I'm adding this param. If we were writing this from scratch I'd say this behavior should be the default but it's turned off by default right now to maintain backwards compatibility. 

**Testing:**
Unit tested. 

**Screenshots (if applicable)**
N/A

**Checklist:**
- [x] All existing tests pass

### DIFF
--- a/docs/bulk_delivery_methods/slack.md
+++ b/docs/bulk_delivery_methods/slack.md
@@ -13,6 +13,7 @@ class CommentNotification
         # ...
       }
     }
+    config.raise_on_failure = true # fail if response is 2xx but body['ok'] is false
   end
 end
 ```
--- a/docs/delivery_methods/slack.md
+++ b/docs/delivery_methods/slack.md
@@ -13,6 +13,7 @@ class CommentNotification
         # ...
       }
     }
+    config.raise_on_failure = true # fail if response is 2xx but body['ok'] is false
   end
 end
 ```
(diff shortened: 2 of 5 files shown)


sherifnada:
@excid3 any thoughts on this PR?

excid3:
Doh, Slack's one of those fun APIs!

It may be a slightly breaking change, but my intention was for it to actually raise an error on failure.

sherifnada:
@excid3 lol exactly. Got bit by this so I figured it'd be useful to have in Noticed directly. I _believe_ this should be a backwards-compatible change since the option is set to false by default (the current behavior), or am I misunderstanding how this works? I'll also add a test case that doesn't explicitly set the option to make it super duper clear it's backwards compatible

excid3:
Before you do that, I'm working on a refactor of this real quick. I think this will be good to go shortly.

sherifnada:
@excid3 let me know how I can help, happy to take up any tasks here

PR #524, opened 2025-01-31
Add OneSignal delivery method

Spone:
## Pull Request

**Summary:**
Add a `OneSignal` delivery method

**Related Issue:**
https://github.com/excid3/noticed/discussions/473

**Description:**
We used the Twilio delivery method as a starting point to implement OneSignal.

**Testing:**
We added a tests based on Twilio delivery method tests, but I think more thorough testing would be needed.

**Checklist:**

- [ ] All existing tests pass

**Additional Notes:**

### DIFF
--- a/lib/noticed/delivery_methods/one_signal.rb
+++ b/lib/noticed/delivery_methods/one_signal.rb
@@ -1,0 +1,56 @@
+module Noticed
+  module DeliveryMethods
+    class OneSignal < DeliveryMethod
+      required_options :include_aliases
+
+      def deliver
+        post_request url, headers:, json:
+      rescue Noticed::ResponseUnsuccessful => exception
+        # OneSignal might returns 200 with errors
+        # - "All included devices are not subscribed"
+        # - Invalid aliases
+        if exception.response.code.start_with?("4") && config[:error_handler]
+          notification.instance_exec(exception.response, &config[:error_handler])
+        else
+          raise
+        end
+      end
+
+      def json
+        evaluate_option(:json).merge({app_id:}) || {
+          app_id:,
+          include_aliases:,
+          contents: params.fetch(:contents),
+          target_channel: "push"
+        }
+      end
+
+      def url
+        evaluate_option(:url) || "https://api.onesignal.com/notifications?c=push"
+      end
+
+      def headers
+        {
+          Authorization: api_key,
+          accept: "application/json"
+        }
+      end
+
+      def app_id
+        evaluate_option(:app_id) || credentials.fetch(:app_id)
+      end
+
+      def api_key
+        evaluate_option(:api_key) || credentials.fetch(:api_key)
+      end
+
+      def include_aliases
+        evaluate_option(:include_aliases)
+      end
+
+      def credentials
+        evaluate_option(:credentials) || Rails.application.credentials.one_signal
+      end
+    end
+  end
+end
--- a/test/delivery_methods/one_signal_test.rb
+++ b/test/delivery_methods/one_signal_test.rb
@@ -1,0 +1,75 @@
+require "test_helper"
+
+class OneSignalTest < ActiveSupport::TestCase
+  setup do
+    @delivery_method = Noticed::DeliveryMethods::OneSignal.new
+    @config = {
+      app_id: "123456",
+      api_key: "key",
+      json: -> {
+        {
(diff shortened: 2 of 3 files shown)


lib/noticed/delivery_methods/one_signal.rb
```diff
@@ -0,0 +1,56 @@
+module Noticed
+  module DeliveryMethods
+    class OneSignal < DeliveryMethod
+      required_options :include_aliases
+
+      def deliver
+        post_request url, headers:, json:
+      rescue Noticed::ResponseUnsuccessful => exception
+        # OneSignal might returns 200 with errors
```
Spone, line 9: image

From https://documentation.onesignal.com/reference/push-notification

There's currently no way to handle that because `Noticed::ResponseUnsuccessful` is not raised when the response code is 200.

excid3: You can parse the response and raise the exception if there is an error in the body.

excid3:
I'm not familiar with OneSignal API but looks like they also support email, SMS, so should this be named more specifically for the type or is their API generic enough to support all of their notification types?

Spone:
> I'm not familiar with OneSignal API but looks like they also support email, SMS, so should this be named more specifically for the type or is their API generic enough to support all of their notification types?

I'll look into it. We're only using push in our current project. At first sight, it seems you just need to alter the `target_channel` param.

image

Spone:
Not sure how to tackle the pipeline failures:

```
lib/noticed/delivery_methods/one_signal.rb:7:46: Lint/Syntax: unexpected token tCOMMA (Using Ruby 3.0 parser; configure using `TargetRubyVersion` parameter, under `AllCops`)
  lib/noticed/delivery_methods/one_signal.rb:14:65: Lint/Syntax: unexpected token tCOMMA (Using Ruby 3.0 parser; configure using `TargetRubyVersion` parameter, under `AllCops`)
  lib/noticed/delivery_methods/one_signal.rb:25:46: Lint/Syntax: unexpected token tCOMMA (Using Ruby 3.0 parser; configure using `TargetRubyVersion` parameter, under `AllCops`)
  lib/noticed/delivery_methods/one_signal.rb:28:44: Lint/Syntax: unexpected token tCOMMA (Using Ruby 3.0 parser; configure using `TargetRubyVersion` parameter, under `AllCops`)
  lib/noticed/delivery_methods/one_signal.rb:66:1: Lint/Syntax: unexpected token $end (Using Ruby 3.0 parser; configure using `TargetRubyVersion` parameter, under `AllCops`)
  ```

**Comment by excid3** [comment]
Standard is set to use Ruby 3.0 syntax but not newer (for compatibility) so looks like you're maybe using some newer features?

**Comment by Spone** [comment]
Got it! Not sure what's going on there: https://github.com/excid3/noticed/actions/runs/13077089941/job/36491828610?pr=524

**Comment by Spone** [comment]
@excid3 any clue about the CI fail?

**Comment by excid3** [comment]
@Spone not sure, but you can try merging in main and that should fix it.

File: test/delivery_methods/one_signal_test.rb [hunk 2]
```diff
@@ -0,0 +1,82 @@
...
+        "Content-Type" => "application/json"
+      },
+      body: {
+        include_aliases: {
+          external_id: "1234567890"
+        },
+        contents: "Hello world",
+        target_channel: "push",
+        app_id: "123456"
+      }
```
excid3, line 35:
```suggestion
      }.to_json
```

This should fix that stub.

lib/noticed/delivery_methods/one_signal.rb
```diff
@@ -0,0 +1,69 @@
...
+      rescue Noticed::ResponseUnsuccessful => exception
+        if config[:error_handler]
+          notification.instance_exec(exception.response, &config[:error_handler])
+        else
+          raise
+        end
+      end
+
+      def json
+        evaluate_option(:json).merge({
```
excid3, line 25: This will override the given JSON, so it's probably better to use this?

```suggestion
        evaluate_option(:json).with_defaults({
```

I think the `||` is also unnecessary since won't you always have an object so the `or` would never evaluate?

Spone:
Sorry for not wrapping this up, I moved on to other things... Feel free to reuse this as a starting point for another PR.

PR #574, opened 2026-01-12
Actually drop Rails < 7.0

neilvcarvalho:
## Pull Request

**Summary:**
The v3.0.0 changelog mentions that Rails 6.1 support was dropped, which was done by removing it from the CI matrix. However, there's still some code that checks for versions lower than 7.0, and 6.1 is still included on the gemspec.

**Related Issue:**

**Description:**

This commit addresses that by changing the dependency version, and removing code that isn't needed anymore.

**Checklist:**

- [x] All existing tests pass

**Additional Notes:**

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     noticed (3.0.0)
-      rails (>= 6.1.0)
+      rails (>= 7.0.0)
 
 GEM
   remote: https://rubygems.org/
--- a/app/models/concerns/noticed/deliverable.rb
+++ b/app/models/concerns/noticed/deliverable.rb
@@ -96,16 +96,7 @@ module Noticed
         self.notifications_count = recipients_attributes.size
         save!
 
-        if Rails.gem_version >= Gem::Version.new("7.0.0.alpha1")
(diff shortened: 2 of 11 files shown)


excid3:
Thanks @neilvcarvalho 👍

excid3:
Need to run `bundle exec appraisal update` as well to update the gemfiles for CI to pass.

neilvcarvalho:
@excid3 I pushed an amended commit with the Appraisal updated lockfiles.

Tip: [Appraisal recommends not checking in the lockfiles](https://github.com/thoughtbot/appraisal?tab=readme-ov-file#version-control):
> When using Appraisal, we recommend you check in the Gemfiles that Appraisal generates within the gemfiles directory, but exclude the lockfiles there (*.gemfile.lock.) The Gemfiles are useful when running your tests against a continuous integration server.

That way, each test matrix entry will resolve the version independently, solving some issues where the same gem is resolved to different versions on the same Rails version depending on the Ruby version. I had issues with `factory_bot` because of that and following that recommendation worked like a charm.

excid3:
Appraisal isn't well maintained, so I only use it to generate the Gemfiles. We don't use it for anything else.

2.5.0 was released in July 2023, 3.0.0.rc1 in October 2024 and nothing since.

PR #576, opened 2026-02-23
Fix race condition in job enqueuing by using after_commit

courtneymiller2010:
## Summary
Fixes a race condition where EventJob could execute before the database transaction committed, causing jobs to fail when trying to access notifications that weren't yet visible in the database.

## The Problem
Previously, jobs were enqueued immediately after saving events and notifications, but before the transaction committed. This created a race condition where:
- The job could start executing before data was committed
- If a transaction rolled back, the job would still be enqueued (incorrect behavior)
- Notifications might not be visible to the job

## The Solution
- Move job enqueuing inside the transaction block
- Use `ActiveRecord::Base.current_transaction.after_commit` to defer job enqueuing until after the transaction commits
- Add support for Rails 7.2+ automatic job deferral while maintaining backward compatibility
- Jobs are now only enqueued if ALL transactions (including user transactions) commit successfully

## Transaction Safety
The `deliver` method creates its own transaction internally. When called inside a user's transaction, this creates a nested transaction. The `after_commit` callback correctly waits for the outermost transaction to commit, ensuring:
- Jobs only enqueue if all transactions succeed
- Events, notifications, and jobs are all rolled back together if any transaction fails
- Safe to call `deliver` inside user transactions

## Changes
- **app/models/concerns/noticed/deliverable.rb**: 
  - Move job enqueuing inside transaction with `after_commit`
  - Add `will_defer_job_enqueuing?` helper for Rails 7.2+ compatibility
  - Add comprehensive documentation on transaction behavior
- **test/models/noticed/event_test.rb**: 
  - Add test isolation with setup block
  - Add tests for transaction rollback scenarios
  - Add tests for nested transaction handling
  - Add test verifying notifications are accessible when jobs run
- **CHANGELOG.md**: Document the bugfix

## Test Coverage
All tests pass (10 tests, 27 assertions):
- ✅ Transaction rollback prevents job enqueuing
- ✅ Nested transaction rollback prevents job enqueuing
- ✅ Notifications are accessible when jobs execute
- ✅ Works with both Rails automatic and manual deferral

## Backward Compatibility
This change is backward compatible:
- Works with Rails 6.1+ through Rails 8.x
- Automatically uses Rails 7.2+ automatic deferral when available
- Falls back to manual `after_commit` for older Rails versions
- API remains unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)


app/models/concerns/noticed/deliverable.rb
excid3, line 127: This doesn't seem correct as the Event and it's notification records are all saved in this transaction.

excid3: Or are you saying if the noticed delivery is inside an external transaction?

courtneymiller2010: The race condition exists even without an external transaction.

The current code enqueued the job immediately after the transaction block closed (line 112). However, closing the transaction block doesn't guarantee the transaction has fully committed to the database. The job could be enqueued and start executing before the transaction is visible to the job worker's database connection, causing the job to fail when it can't find the records.

The fix ensures jobs are only enqueued after the transaction actually commits using after_commit. For Rails 7.2+, we can rely on `enqueue_after_transaction_commit` to handle this automatically. For earlier versions, we manually defer enqueuing with current_transaction.after_commit.

These blog posts explain the issue and solution in more detail:
  - https://prateekcodes.com/rails-72-enqueue-after-transaction-commit/
  - https://prateekcodes.com/rails-82-enqueue-after-transaction-commit-default/

courtneymiller2010:
After further investigation, I'm closing this PR as I no longer believe it addresses the actual root cause.

We initially suspected a race condition between the database transaction and job enqueuing: specifically, that the job was being enqueued before the transaction committed, causing the async worker to query records that didn't exist yet. We internally added a 5-second wait as a workaround and pushed up this PR as a longer term fix.
                                                                                                                                                                                                                                        
After taking some time away from this issue and re-examining the source code, we agree it already handles this correctly by deferring enqueuing until after the transaction completes. So the race condition we theorized shouldn't be possible with the current implementation.                                                                                                                                                                                            

The outage symptoms we saw (notifications not existing when the job ran) must have a different root cause that we haven't identified yet. We're pulling back this change to avoid introducing unnecessary complexity while we continue investigating.

Thanks for your time, and apologies for the noise.

